PR #91, opened 2019-09-24
Add Bootloader...

WestfW:
I think this is ready for at least a try.
Things to think about:

1. I currently put in a "bootload" sub-menu in each of the old board types, but it sort-of gets lost with all the other sub-menus.  So I also added a set of new "boards" named "<oldname> (optiboot)"  I'm not sure that having both is a great idea.

1. I changed the behavior of "burn bootloader" pretty significantly.  In you current code, this burns ALL the fuses as per the menu settings (and doesn't actually upload a bootloader.)  In my new version, "burn bootloader" ONLY changes the BOOTEND fuse and uploads Optiboot.   To program the other fuses, you'd need to use "upload using programmer" from one of the non-optiboot boards

1. I'm pretty sure that "upload using programmer" does NOT work right if you've selected an optiboot board configuration.  I think it still works on the non-optiboot boards.


SpenceKonde:
Thanks, looking forward to playing with this tonight!

May I ask what the reasoning behind making burn bootloader not set other fuses was? Is there some complication I'm not seeing here?

WestfW:
I think that I was thinking that the Fuses are much more varied and independent than old parts, so a "set up for using bootloader" command should only modify the (one) fuse needed to enable the bootloader.  Thinking again, it's also the "last chance" a user has to define fuses for the to-be-bootloaded system, so it probably makes more sense that it can program all of them.

The optiboot_x documentation is lagging behind.  The general build command is:
```
make -f Makefile.mega0 <boardtarget>
make -f Makefile.mega0 <cputarget> <options>
ie:
make -f Makefile.mega0 xplained817
make -f Makefile.mega0 drazzy412
make -f Makefile.mega0 attiny416 LED=A7 UARTTX=B2 TIMEOUT=1 BAUD_RATE=57600
```
I guess that `make <boardtarget> <overridingOptions>` should work, but I'm pretty sure it doesn't.

Feel free to submit issues to the Optiboot repository, or to my fork of megaTinyCore, for anything you want me to try to deal with.  I'm a big believer in using issues to track ... well, **issues** ... rather than some sort of Hallowed ground where you can only add things that have gone through some sort of confirmation process because a bug count might affect someone's performance review...

PR #112, opened 2019-11-15
Add printf to print class

MCUdude:
Printf is not a part of the Print class, and it seems like it's going to stay this way in the official Arduino cores. However, this doesn't mean our cores cannot support it!

Discussion about printf here: https://github.com/arduino/ArduinoCore-API/pull/28


MCUdude:
It may be a good idea to add a small note in the README that printf is available for use. Most users probably don't know this. [Here's my take on it](https://github.com/MCUdude/MegaCoreX#printf-support).

PR #122, opened 2019-12-03
Update README.md

avandalen:
Small change of the text.


SpenceKonde:
Thanks for reporting - I expanded that section, but went a bit further to capture all the ramifications of the autoreset circuit.

PR #123, opened 2019-12-13
Fix max/min PWM value

MCUdude:
The output of a PWM pin should be completely low when a 0 is passed to the analogWrite function. It should also be completely high when 255 is passed


SpenceKonde:
Hmm, can you explain the 

#define PWM_TIMER_PERIOD	0xFE	/* For frequency */

Aren't there still 255 (0xFF) ticks in a timer period?

MCUdude:
I just applied the patch I was "given" a while ago.
https://github.com/MCUdude/MegaCoreX/pull/53

It never caused any issues that I'm aware of, but I'll dive into the datasheet just to be sure. OTOH @egilkv has proven himself many times to be pretty skilled, so I'm sure it has a purpose.

SpenceKonde:
Ooooooooh! I get it now!

SpenceKonde:
Thanks!

MCUdude:
It should probably be added to the official repo as well, but I haven't really bothered.

Issue #54, opened 2019-06-25
Reduce flash/memory use of core and libraries


SpenceKonde:
Current the core and it's libraries are flash hogs. This is problematic on the smaller ATtiny parts. We now create the assembler listing when dumping the compiled binary, so we have the tools we need to see where space is going and whether we can slim things down. 

It would be helpful to flag up any libraries or core functions that you find to be particular space hogs for special attention within this issue.

kadamski:
In my case (attiny416), just adding Wire.h as an include adds ~2300 bytes to the size of a program. If slave support is removed, this only adds ~1400 bytes so maybe it would make sense to either split master and slave implementation or make it possible to remove one?

Removing the possibility to change baud rate at runtime saves about 150 bytes. If we also constify the F_CPU_CORRECTED (which is currently always equal to F_CPU), another 250 bytes is saved.

kadamski:
BTW, we don't have to remove whole Slave functionality - we would just have to remove the Slave ISR (TWI0_TWIS_vect) as it always calls TWI_SlaveInterruptHandler() which brings some more Slave related funcitons as dependencies which cannot be removed from the binary. So an ifdef that enables/disables Slave/Master ISR could be useful.

SpenceKonde:
Serial can't get the same treatment without more careful investigation; based on discussion in #130 I am pretty sure this will break stuff, but that it could probably be bullied into working in most use cases. @kadamski were you able to make changes for Serial and compile sketches that used serial after doing so?

saysphilippe:
Hi! I'm trying to use the SoftwareSerial towards the Attiny202 but getting a compile error. When I deploy the same code to 1614 or 1604 there is no problem. I'm trying to use a 202 or 402 with a HC-12 to send a message. My application: Mailbox notifier. :)
The error message says:

[...]

PR #129, opened 2019-12-20
twi: operate on khz

kadamski:
Operation on 32-bit values takes a lot of instructions on our 8bit CPUs.
In this case this can be avoided as we know we are only operating on khz
values anyways so we can pre-compute it. The value in khz fits nicely in
16-bit so the generated code will be more compact. As an added bonus I
think it also makes those values easier to read (you don't have to count
those zeros :)).

This reduced size of my sketch from 2926 to 2828 (98 bytes) on my t416.

This is related to #54.


SpenceKonde:
This seems reasonable - if it reduces the sketch size, I see no problem with this.

PR #130, opened 2019-12-20
TwoWire: avoid vtable by not ineriting HardwareI2C

kadamski:
In an attempt to reduce the size or our binaries, we use --gc-sections
linker functionality in hopes not used functions will be automatically
removed by the linker. If, however, you check the symbols in the out ELF
file when Wire.h is included, you can find that all of the methods of
TwoWire class (even those rarely used) were still there.

The linker cannot do that because TwoWire is inheriting from HardwarI2C
class which in turn defines all of the methods as virtual. This means a
vtable is created and it references all of the methods preventing linker
from removing them.

By giving up on this pretty OO concept (which I think is not often used
in Arduino world) we can reduce the size of the binary in many cases.
This change reduces the size of an empty sketch including Wire.h from
2664 bytes to 1478 bytes.

This is related to #54.

### DIFF
--- a/megaavr/libraries/Wire/src/Wire.h
+++ b/megaavr/libraries/Wire/src/Wire.h
@@ -32,7 +32,7 @@
 // WIRE_HAS_END means Wire has end()
 #define WIRE_HAS_END 1
 
-class TwoWire : public HardwareI2C
+class TwoWire : public Stream
 {
   private:
     static uint8_t rxBuffer[];


kadamski:
@SpenceKonde thich change might be more controversial than others I did but it also gives biggest benefits so far so I think it's worth considering. Maybe some other way to avoid having vtables could be used (to be honest, I'm not a C++ guy).

SpenceKonde:
Wow, what a huge savings in space. This is the sort of thing we really need to make the smaller parts more usable. 

I'm wondering if there are any cases where this could cause problems, though? I don't really know much about vtables either....

SpenceKonde:
I'm particularly worried about whether there are libraries that expect TwoWire to be a subclass of HardwareI2C....

kadamski:
I think this is the only problem that we could encounter - if there is some code that would like to take a pointer to any HardwareI2C class. And now this is a though question - if you have 16kb of flash or more, you may be ok with such an overhead for this library. If you only have 2-4kb, though, you may not have enough space for this library anyways as the Wire library is taking most of that already itself.

One option would be to make this change "reversible" by some #define. Or the other way around, we could have an opt-in define like #ifdef WIRE_COMPACT. Anybody interested more in the flash space than compatibily could define this before including Wire.h (or other libraries which could have "compact" version as well).

Another option is to do this automatically.. like do an arbitrary decision that we will have this optimisation on 2-4kb parts but not in larger ones

I think the "opt-in" option might be the most conservative. Either way, is any if above is choosen, this should be documented so let me know what you think and I will repost with a proper additions to the readme.

SpenceKonde:
I think the WIRE_COMPACT approach is best. If you adjust your PR to #ifdef it with WIRE_COMPACT, I will merge it and do the hocus-pocus in the board menu to support it; I will default to the compact version unless someone finds in testing that there are a lot of popular libraries that fail with it.

SpenceKonde:
I added the menu option. Merging and adding the #ifdef now.

SpenceKonde:
Okay, this is in. We should be on watch to see how many I2C device libraries barf on this, 

I'm wondering if we could instead modify the HardwareI2C library so that those functions are not virtual - since as I understand this is only necessary if an instance of a subclass of HardwareI2C is defined at runtime, which Wire is not...

kadamski:
@SpenceKonde Thanks for taking care of that, I had a lot going on recently and could not work on that.

If the HardwareI2C methods are declared as virtual, you can do something like that:
```
#include <Wire.h>

void setup() {
  HardwareI2C *w = &Wire;
  w->begin();
}
```
And this will call begin() from TwoWire class, thanks to the generated vtable. If, however, the "begin" method was not virtual, this would try to call the begin implementation of the HarwareI2C (and we don't have it).

The compact version as we have now will not compile because the TwoWire is not inheriting from HardwareI2C anymore so this cast of pointers will not work at all.

So we can't give up on using virtual functions if we want to support such casts. What we could (I think) do, would be to just get rid of the TwoWire class and implement everything in the HardwareI2C class directly, with no inheritance. Then, only HarwareI2C class would exist, and it would be possible to take a pointer/reference to this class. There will be a problem, however, if somebody would try to take a pointer/instance for a TwoWire class in this case, however. To fix that, we could do an alias using "using" keyword (or typedef), though. This way TwoWire and HardwareI2C will basically mean the same. 

What do you think?

kadamski:
BTW, taking into account how inheritance is done in the Arduino, I don't think there is a lot code using it. Why would somebody want to take a pointer/reference to HardwareI2C instead of TwoWire for example? I don't know. It would make sense if there were two implementations - like HardwareI2C and SoftwareI2C and I would like my code to work with both of them but they would need a common ancestor, like GenericI2C and this is not the case in Arduino. So the only option would be to have a custom implementation of the TwoWire, which would also inherit from HardwareI2C.. but I don't see any reason why somebody would want to do that.

This would make more sense for the serial as it is common to use SoftwareSerial instead of HardwareSerial in Arduino world. But also they do not share common base class (other than Stream) so if you want to be able to handle any of the Software or Hardware serial implementation, you have to work with Stream anyways. There's 7 years old issue open for that on github and it isn't changing (https://github.com/arduino/ArduinoCore-avr/issues/228)

So I don't see much practical sense in making HardwareSerial / HardwareI2C an abstract class as it is in megaavr. And it means more flash space used which in case of tinys is an important resource.

SpenceKonde:
hmm, well - for such cases, there's always the option of using the non-compact option. 

The real question IMO is whether there are sketches/libraries in the wild which do this, and if so how common they are. If they exist in the wild and are common, well, then this approach would probably not be any good. So far I have found one - at least one HMC5883 library

SpenceKonde:
I wonder if this is part of the reason why so many basic sketches end up unreasonably large

SpenceKonde:
AAHAAAA! This is not an issue for code in the wild! Why did I not think to check what ATTinyCore does? It just has it as a stream. Nobody has objected. Hence, the new menu option and documentation for it will be removed before 1.1.3 is released.

kadamski:
Well, as I said the inheritance in Arduino is broken anyways so I doubt there is much code using it. And our tinies are already quite restricted so I think it just makes sense to make 99% of.sketches smaller even if it breaks this 1% of cases somehow.

But your observation based on normal AttinyCore is for I2C.. how about Serial ?

PR #139, opened 2020-01-29
Documentation improvements

veganism:
The included guides are **extremely** helpful and include tons of information on using this core and programming in general - however, I noticed a few glaring typos, so I wanted to fix those. In the process of reading everything, I went ahead and corrected every mistake I could find, and did some minor editing to (hopefully) improve most of the documentation. This includes changing the markdown for code examples where applicable, correct usage of "it's" and "its," and other changes. It's a lot to sift through, so I tried to stay as conservative with the changes as I could while still making sense.


SpenceKonde:
Wow, this is wonderful! Thank you!!

PR #146, opened 2020-02-22
Adding the ability to receive broadcast messages in Wire lib

LordJakson:


LordJakson:
The line 

	TWI0.SADDR = (address << 1) | (receive_broadcast & 1);

is even better

SpenceKonde:
Eons later, I finally figured what this is supposed to do, and recognize that it makes sense. Thanks for the PR. While I'm looking at it, I'm going to expose a few other features of the underlying hardware too via additional parameters that can be optionally passed to begin() along with the slave address,

PR #148, opened 2020-03-03
Updated programmers.txt to support Atmel ICE UPDI

rneurink:
Added support for the Atmel ICE UPDI programmer.
Be aware that the Atmel ICE might require a firmware update as it needs V1.36!! This can be done with the Atmel studio.

Have tested this and it works. 
This is mainly an easy way to program the explained pro kits with an Atmel ICE programmer.
[Microchip docs](https://microchipdeveloper.com/atmelice:updi)
afbeelding
afbeelding


SpenceKonde:
Nice, thanks!

PR #149, opened 2020-03-03
Updated programmers.txt to support EDBG for explained pro

rneurink:
Added support for the Atmel embedded debugger on the explained pro kits

Have also tested and it works flawlessly as AVRdude config already contains everything for this


SpenceKonde:
Thanks, not sure how I missed this earlier

PR #233, opened 2020-09-28
fix do_nvmctrl()

WestfW:
tested on Curiosity Nano ATtiny3217.
.hex files compared to orignals to confirm they only differ in the relevant area.


WestfW:
I left your reset cause edits in there, so you don't quite match the current optiboot repository.
Look at incorporating your changes in the main branch...

SpenceKonde:
Thanks - I may yet do more work on reset cause; not particularly confident in it currently.

PR #235, opened 2020-10-04
TWI_SlaveAddressMatchHandler interrupt timing issue fix

raspy135:
I'm new to AVR core so I could be wrong, although I noticed some unstable behavior with Master read operation.

ATTINY1607(Slave)
ESP32(Master)
I2C rate is 400kHz
16Mhz CPU clock.

Slave(ATTINY) sends 4 bytes. I observed about 1 to 5% read error on Master side.

Enabling Data interrupt in TWI_SlaveAddressMatchHandler() is a bit too early, sometimes interrupt comes earlier than data write to the txBuffer in user handler.

By this change I could get rid of the error.


SpenceKonde:
Didn't see this in time for 2.1.0, and there's a merge conflict that needs to be resolved, so 2.1.1

SpenceKonde:
Thanks - that change looks reasonable to me.... and if you've got a test case that demonstrates the improvement in behavior *shrug*... As I recall, I2C implementations on the ESP's often do not subscribe to a literal interpretation of the I2C specification (they don't even have it implemented in hardware do they? I have seen some horrifying implementations that the authors called I2C for the 8266)

PR #276, opened 2020-12-10
Fix broken condition

yumkam:
Regression by
a804e54ce79ad651e87445db458d12404bf6680d (recent formatting fix)


SpenceKonde:
YIKESSSSS!!! Thanks!!!

PR #318, opened 2021-02-12
Add Serial1 support to 2-series

tagno25:
This enables Serial1(USART1) on 2-series, but pin swap is not working for either port.


SpenceKonde:
Splendid! Thanks! (also, hot damn, this is the first time I've seen the github action bot pop in with info, i love it!)

I really ought to fix those example errors.

SpenceKonde:
CI now passes so when you submit PR's you no longer get emails about preexisting errors. And if your PR adds new errors , well, you deserve the emails! :-)

SpenceKonde:
(also, by the way - I changed it to test whether USART1 was defined rather than looking for our define.  always pick defines from headers over ones from third party code, and try to test for peripherals rather than identity - pays off down the line.

PR #329, opened 2021-02-18
Add warning on burning bootloader

amadeuspzs:
The "Serial port and 4.7k (pyupdi style)" method is not reliable as per #320 .

Add language as a warning to other users.


SpenceKonde:
I will try to get this issue fixed asap, have already let the dev know - in the meantime I will add warning that specifically states when it doesn't work.

PR #333, opened 2021-02-23
Remove unnecessary dependence on micros() for neopixel libs

SpenceKonde:
in the cases where we don;t have micros(), we now issue warning, and we now correctly include rtc millis cases, where they were previously not checked for.


SpenceKonde:
Issue raised in #158 - this change is in both branches.

ArnieO:
Tried the update now. 
This seems like an actually missing '('?
image

SpenceKonde:
Yup. Static had an extra close paren, non-static missing one, and I didn't bother to check the CI results. Terribad!

PR #353, opened 2021-03-10
Update for PORTA ISR

dpnebert:
PORTA ISR was setting flags equal to INTFLAGS, and then INTFLAGS equal to flags, never setting the bit for the respective pin to 1.

I got started on the series-0 using megaTinyCode, so thank you for the giant boost in my education.

### DIFF
--- a/megaavr/extras/PinInterrupts.md
+++ b/megaavr/extras/PinInterrupts.md
@@ -83,7 +83,7 @@ void loop() {
 
 ISR(PORTA_PORT_vect) {
   byte flags=PORTA.INTFLAGS;
-  PORTA.INTFLAGS=flags; //clear flags
+  PORTA.INTFLAG=0xff // clear all PORTA flags
   if (flags&0x02) {
     interrupt1=1;
   }


SpenceKonde:
I really appreciate getting PR's, so thanks for this - but I don't think your suggestion works here - there's a blatant bug that is just a typo, and a much more subtle one...

First off, your PR refers to PORTA.INTFLAG,. The register is named PORTA.INTFLAGS.

Secondly, why is 
```
  byte flags=PORTA.INTFLAGS;
  PORTA.INTFLAGS=flags; 
```

not sufficient here? Whichever bits are set, indicating an interrupt condition has occurred will be 1 in flags. So when we write that back to INTFLAGS, that should clear all the bits that are set... So does yours - except for one thing: there's a  subtle race condition with multiple interrupt conditions on a single port: if one of them occurs at the exact moment that the ISR is executing because of signal on another pin within that port, and the interrupt condition is met between t  `byte flags=PORTA.INTFLAGS;` and `PORTA.INTFLAGS=0xFF;` then you could read the flags *without* the new interrupt''s having happened recorded, then it happens, and then you clear PORTA.INTFLAGS by writing 0xFF to it - now all record of that having happened have been 'black hole'ed' - the code would never see that it happened and couldn't react to it! Using the method I suggest, in that case, since we didn't clear the bit that we didn't know was set, the bit would still be set at the end of the interrupt, and so the ISR would immediately fire again and could handle the sneaky event.

Am I missing something?

dpnebert:
Hi there, sorry for the delay.  I saw the notification on Thursday and re-read the message several times wanting to make sure I completely understood what you were saying.

To address the typo.  I kept changing what I was suggesting and at the end, I didn't go back and check, but thank you for continuing on with the rest of the explanation.  I'm still learning much of this and it helped to have more information.  So thank you for spending the time to respond back.

I want to make sure I understand this sentence, "Whichever bits are set, indicating an interrupt condition has occurred will be 1 in flags."  I read that to be, "Whichever bits are set, ... will be 1 in flags.

I am working with my new ATTINY1627 curiosity board, and I want to paste in a section so that you can see what I am reading on page 156, 17.5.10 Interrupt Flags: 
```
Pin interrupt flag n is cleared by writing a ‘1’ to it.
Pin interrupt flag n is set when the change or state of pin n (Pxn) matches the pin's Input/Sense Configuration (ISC)
in PORTx.PINnCTRL.
Writing a ‘0’ to bit n in this bit field has no effect.
Writing a ‘1’ to bit n in this bit field will clear Pin interrupt flag n.
```

Ok, so assuming (only) PA0's interrupt was triggered, INTFLAGS would be equals to "0x11111110", which means to set a bit, would be it was '0'.
And if we assign flags to be equal to "0x11111110"

Am I on track so far?

SpenceKonde:
Sorry, missed this last week (where did the last 7 days go?!)
AVR interrupt flags are somewhat strange....

They read 0 when not set, and 1 when set. 

So say I enable pin interrupts on PA0, PA1, and PA2 We start off with `PORTA.INTFLAGS=0b 0000 0000` (spaces just to make it easier to read). Something flips PA0. Now  `PORTA.INTFLAGS=0b 0000 0001` and the ISR is called. We need to write `0b 0000 0001` to PORTA.INTFLAGS` in the ISR to clear the flags (otherwise the moment the ISR finishes, the CPU will be like "okay, we're out of the ISR, lets see, do we have any interrupts to handle? Ohg, yes we do - PORTA.INTFLAGS is non-zero, call the PORTA interrupt!" - a situation that I've wound up in a number of times). 

And yes that means that if PORTA.INTFLAGS starts off with 0b 0000 0001, then we write 0b 0000 0001 to it, it will no longer be equal to that value. So yeah - that's why I write INTFLAGS back to it. 

There is no way to set a bit in the INTFLAGS registers to 1 other than triggering the interrupt condition; bits that are 0's are ignored, and bits that are 1's are cleared. 

One thing worth noting is that one can imagine an ISR with a few sections, each of which covers a different interrupt conditions.... The ISR might choose to set the bit for the pin it's doing things trihgered by while leaving the others undisturbed and return from the interrupt. This would allow one to check if other interrupts need to be serviced, for example. 

Also, if you are clearing bits one at a time, you can, and probably should, do `VPORTA.INTFLAGS |= 1<<bit_pos`

dpnebert:
_Sorry, missed this last week (where did the last 7 days go?!)
AVR interrupt flags are somewhat strange...._

No worries, I just kept checking to see if the PR was still open and waited. :)

_They read 0 when not set, and 1 when set._

I think this is one of the disconnects in understanding we have together.  In the ATTINY1627 datasheet, on page 156 for interrupts, it says,

```
Writing a ‘0’ to bit n in this bit field has no effect.
Writing a ‘1’ to bit n in this bit field will clear Pin interrupt flag n.
```

Which I guess can be inverted, but for the sake of learning, I'll roll with your theory of '0' is cleared and '1' is set.

_So say I enable pin interrupts on PA0, PA1, and PA2 We start off with PORTA.INTFLAGS=0b 0000 0000 (spaces just to make it easier to read). Something flips PA0. Now PORTA.INTFLAGS=0b 0000 0001 and the ISR is called. We need to write 0b 0000 0001 to PORTA.INTFLAGS` in the ISR to clear the flags (otherwise the moment the ISR finishes, the CPU will be like "okay, we're out of the ISR, lets see, do we have any interrupts to handle? Ohg, yes we do - PORTA.INTFLAGS is non-zero, call the PORTA interrupt!" - a situation that I've wound up in a number of times)._

I believe you meant the above paragraph that I need to clear the interrupt flag in the ISR, or else it will continue to fire of the ISR because the flag is never cleared.

I also believe you had a typo that is really confusing me:

`We need to write 0b 0000 0001 to PORTA.INTFLAGS` in the ISR to clear the flags`

If we want to clear the flag, don't we write a '0' for PA0?

_And yes that means that if PORTA.INTFLAGS starts off with 0b 0000 0001, then we write 0b 0000 0001 to it, it will no longer be equal to that value. So yeah - that's why I write INTFLAGS back to it._

Sorry, I really trying to read this and fully understand it, but I don't know what you mean by, "it will lo longer be equal to that value."  If the ISR is called because PA0 flips then,

```
byte flags = PORTA.INTFLAGS;      // flags now equals 0b 0000 0001
PORTA.INTFLAGS=flags;             // INTFLAGS now equals 0b 0000 0001
                                  // INTFLAGS still non-zero and the ISR will fire again
```

SpenceKonde:
No, you don't seem to be getting the point I am trying again and again to drive home

```
byte flags = PORTA.INTFLAGS; 
// flags now equals 0b 0000 0001
PORTA.INTFLAGS=flags; 
// INTFLAGS now equals 0b 0000 0000 (unless another interrupt condition on some pin other than PA0 between when we read PORTA.INTFLAGS and stored it as flags, and this line caused another bit to be set to 1, in which case that bit would be set, but bit 0 would still read 0 -. 
```
We wrote 0b0000 0001 to it, and per datasheet, writing a 1 to a bit clears it. (a bit is "set" when it is 1 and "clear" when it is 0). Writing 0 to a bit does not change that bit, whether it was a 1 or a 0 before, and writing a 1 to the bit clears it if it was previously set.  
Peripheral registers are not variables. Assigning a value to them does not, necessarily, set the to that value, and what you read from them may or may not have any relation to what you wrote to them last time.. They can do whatever the designer wanted them do when designing it. They are declared as (and the compiler thus treats them as) a volatile variable, since the same set of restrictions that it implies are also sufficient for registers


PORTx.OUTCLR, for example, behaves... actually very similarly. It reads as PORTx.OUT, but when you write to it, clears every bit written as a 1. So if someone were to try switch a pin from HIGH to LOW with PORTA.OUTCLR |= (1<<bit_to_clear)... that would be a bug, as it would do a read-modify-write, so it would get the pins currently set high, OR that with the bit they wanted to clear, (likely yielding the same value) and then write it back - it would thus set all pins on the port LOW! The |= 1<<bit ideom is so common in embedded programming that I read a dude's code like 4 times last time it came up before I realized that was what his problem was.

Every time I help someone who got caught by that I try to think of a way to make the core conspire with the compiler to error out when someone does that. Haven't been able to come up with a viable approach short of an external script to grep the sketch for such assignments.That is a can of worms I ain't gonna open (unless, of course, Microchip or someone else. offers to give me some fat stacks 💰  to make it happen)

SpenceKonde:
The case I mention in the code block where the ISR is called, you read the INTFLAGS register, and write it back to clear the bits you've read in - but another interrupt condition happens between when you read the INTFLAGS and when you write back to it, is actually super important, and is the reason they have this wacky behacvior on the INTFLAGS registers. I don't think there is any other behavior (besides the -0 is ignored 1 clears the current value) possible for a register with more than 1 interrupt flag in it that will not lose interrupts i that case.

SpenceKonde:
no reply for over a week-> closing w/out merge.

PR #398, opened 2021-04-23
Fix path to PACKS

kroesche:
I had to make the following change in order to be able to use PACKS. I'm not sure if I was doing something wrong but when I used PACKS=path/to/packdir on my build command line, the -B path was messed up used the make target name instead of the device name.

### DIFF
--- a/megaavr/bootloaders/optiboot_x/Makefile
+++ b/megaavr/bootloaders/optiboot_x/Makefile
@@ -85,7 +85,7 @@ endif
 
 # If we have a PACKS directory specified, we should use it...
 ifdef PACKS
-PACK_OPT= -I $(PACKS)/include/ -B $(PACKS)/gcc/dev/$*
+PACK_OPT= -I $(PACKS)/include/ -B $(PACKS)/gcc/dev/$(TARGET)
 ifndef PRODUCTION
 $(info   and Chip-defining PACKS at ${PACKS})
 endif


SpenceKonde:
Okay sure I'll let this in. It has no relevance to how I build Optiboot and this makefile shit is way over my head. 

I build optiboot like this, which I'm sure will nauseate anyone who knows how to work with makefiles.: 
I have a 1.0.6 installation of Arduino, I deleted everything except hardware/arduino/bootloaders and hardware/arduino/tools. Replaced hardware/arduino/tools/avr with the new toolchain package (the one you get when you install my cores)
Thus, omake points to the right place and I can use it to build the bootloaders :-) (running windows, if that wasn't clear)

kroesche:
Okay, well I hope that was a good fix. It fixed my problem. I am not working in the Arduino ecosystem. I have a '1614 project and I use Optiboot for the boot loader but not any other Arduino stuff. I have the AVR toolchain from microchip installed on my system and the relevant PACKS directories added to support my devices. I was trying to build Optiboot for my system and I ran into this problem and thought I would submit the PR in case it helps others. Regards. Thanks for maintaining all this.

SpenceKonde:
Woah, I've never heard of using Optiboot completely outside the Arduino environment. Which bootloader entry conditions do you use? do you do the UPDI as reset to enter the bootloader on reset? I always found the 8-seconds a bit fiddly with timing - but I guess that's not a problems outside the Arduino environment (one of the weaknesses of Arduino IDE is you can't upload without it doing a recompile, which makes timing it challenging.

kroesche:
I have the boot loader always start first with a 4 second timeout. Then I can start a firmware update with a power cycle or a reset. My project used an '841 and I was using Optiboot from ATTinyCore (with some mods). I just switched to using a '1614 and I dont have Optiboot working yet. I am using UPDI for programming the boot loader and I dont know what issues I will have if any with the shared UPDI/reset thing. My application can also enter the boot loader so that is the normal way I do updates in the field.

SpenceKonde:
Huh, well, it should work out of the box (though I can't say anything about your builds of it, the binaries I supply most certainly do work. 

What are the symptoms and what build options are you using? After testing and making it work with bootloader binary from megaTinyCore package,. I'd probably build with something like this (note that all tinyAVR 0/1-series parts with more than 8 pins can use the same optiboot binaries, assuming you don't try to do stuff that uses pins that don't exist on the lower-pincount devices; as I understand it, the die is the same for 1614, 1616, and 1617). Do you get the triple blink? What mods, if any, have you tried to carry over?

HELPTEXT += "Generic 14~24 pin dev board, default serial pins, 4 second timeout, LED on A7\n"
tinyxyz4sec: TARGET = attiny3217
tinyxyz4sec:
	$(MAKE) -f $(MF) optiboot_txyz_4sec.hex UARTTX=B2 TIMEOUT=4 LED=A7

	
For running optiboot from within the application, perform a software reset ( `_PROTECTED_WRITE(RSTCTRL_SWRR,1)`)
Optiboot has no guarantee of working when jumped to direct from the app. I have it detect and not jump to app (running bootloader instead) after a direct jump to bootloader (no reset flags set) and I thought I was going to make it do a software reset itself in that case to get back into a known state from which the bootloader can be expected to perform reliably... but I don't see any trace of that part! Not sure if I forgot about it (maybe it was while working on Dx-series that I realized bootloader can malfunction if jumped to directly, I might not have circled back to the tinyAVR version.

kroesche:
I have custom hardware that uses half duplex serial so I need to modify the boot loader to work with half duplex. It is similar to the way the RS485 option worked for the ATTinyCore Optiboot, except there is no turnaround signal. In the case of the new '1614 design, I didn't see anything in Optiboot for RS485 so I will need to modify it for half duplex. With regard to invoking BL from app, for '841 I had to reset some registers first before calling directly to BL and that worked okay. For the '1614 it has a software reset so I am planning to use that.

SpenceKonde:
The hardware has some features to help you on that - take a look at datasheet section 24.3.3.2.6: Half Duplex Operation on page 326 - but it might be hard to make Optiboot work with that. It's too bad you don't get a direction pin; it's got support for either kind of direction pin - either the input where it won't transmit until it gets the right level on that pin, or

PR #422, opened 2021-05-12
Changed TWI baud calculation

rneurink:
Hi @SpenceKonde,

I just saw the latest release notes and saw something regarding the I2C clock (baud rate) and thought I'll post my currently working solution for the attinyxx17 series.
I've been messing around with I2C on the attiny817, attiny1617, and the attiny3217 and I came up with the following calculations.
(F_CPU/(2*frequency)) - (5 + (((F_CPU / 1000000) * t_rise) / 2000)) to calculate the baud rate with t_rise (max) in ns and the frequencies in Hz. This calculation comes from the datasheet for the ATtinyXX17 series. Due to this very generic calculation and the nonlinearity in the frequency transmission I've calculated and tested an extra offset for specific frequency ranges.

I calibrated the internal clock before testing and I've verified the frequencies with a high-end scope. Most frequencies seem to work correctly, however, there is a weird section between 400kHz and 900kHz in which the nonlinearity is not presented well enough in the calculations. I've yet to see devices that work on this frequency range so I think this solution is good enough for now.

I've got an Attiny1627 coming soon on which I can test some of these calculations as well.

If anyone can test this code on a few more devices that would be very helpful. 

Btw I've been working a lot with the XX17 series chips and and got a 1627 coming my way so if you ever need some help with a specific part of an implementation feel free to PM me

### DIFF
--- a/megaavr/libraries/Wire/src/utility/twi.c
+++ b/megaavr/libraries/Wire/src/utility/twi.c
@@ -64,7 +64,7 @@ void TWI_MasterInit(uint32_t frequency) {
   if (twi_mode != TWI_MODE_UNKNOWN) {
     return;
   }
-#ifdef PORTMUX_TWIROUTEA
+  #ifdef PORTMUX_TWIROUTEA
   if ((PORTMUX.TWIROUTEA & PORTMUX_TWI0_gm) == PORTMUX_TWI0_ALT2_gc) {
     // make sure we don't get errata'ed - make sure their bits in the
     // PORTx.OUT registers are 0.
@@ -72,8 +72,8 @@ void TWI_MasterInit(uint32_t frequency) {
   } else {
     PORTA.OUTCLR = 0x0C; // bits 2 and 3
   }
-#else // megaTinyCore
-#if defined(PORTMUX_TWI0_bm)
+  #else // megaTinyCore
+  #if defined(PORTMUX_TWI0_bm)
   if ((PORTMUX.CTRLB & PORTMUX_TWI0_bm)) {
     // make sure we don't get errata'ed - make sure their bits in the
     // PORTx.OUT registers are 0.
@@ -81,12 +81,12 @@ void TWI_MasterInit(uint32_t frequency) {
   } else {
     PORTB.OUTCLR = 0x03; // else PB0, PB1
   }
-#elif defined(__AVR_ATtinyxy2__)
+  #elif defined(__AVR_ATtinyxy2__)
   PORTA.OUTCLR = 0x06; // 8-pin parts always have it on PA1/2
-#else
+  #else
   PORTB.OUTCLR = 0x03; // else, zero series, no remapping, it's on PB0, PB1
-#endif
-#endif
+  #endif
+  #endif
 
   twi_mode = TWI_MODE_MASTER;
 
@@ -115,7 +115,7 @@ void TWI_SlaveInit(uint8_t address, uint8_t receive_broadcast, uint8_t second_ad
     return;
   }
 
-#ifdef PORTMUX_TWIROUTEA
+  #ifdef PORTMUX_TWIROUTEA
   if ((PORTMUX.TWIROUTEA & PORTMUX_TWI0_gm) == PORTMUX_TWI0_ALT2_gc) {
     // make sure we don't get errata'ed - make sure their bits in the
     // PORTx.OUT registers are 0.
@@ -123,8 +123,8 @@ void TWI_SlaveInit(uint8_t address, uint8_t receive_broadcast, uint8_t second_ad
   } else {
     PORTA.OUTCLR = 0x0C; // bits 2 and 3
   }
-#else // megaTinyCore
-#if defined(PORTMUX_TWI0_bm)
+  #else // megaTinyCore
+  #if defined(PORTMUX_TWI0_bm)
   if ((PORTMUX.CTRLB & PORTMUX_TWI0_bm)) {
     // make sure we don't get errata'ed - make sure their bits in the
     // PORTx.OUT registers are 0.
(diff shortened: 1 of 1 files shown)


SpenceKonde:
God, isn't that style checker a pain in the ass sometimes? 
I like some things (the spaces between numbers and operators), and catching deviant submissions where the person puts a newline between the condition of an if () and the opening brace.  
But it's attitude towards indentation is not ideal. I very much wish I could tell it to expect an extra level of indentation for ALL #if conditionals. not just some of the, (I don't even know it's algorithm). There are a lot of files where I wasn't willing to live with it;s formatting demands and disabled indent checking for some or all of a file. 


Huh, you've checked this for safety against baud rate overflows overflows in intermediate calculations and integer math problems? 

That was what was plaguing the past one of the older implementations. that we had - under various circumstances, they would encounter an integer overflow and produce wacky numbers, which is what prompted my solution of sorting it into buckets and using fixed values for each one. But if your solution actually works, that looks a lot more graceful than my hackjob!

Oh - also - what happens if the user requests a speed faster than the hardware is capable of? IIRC in the implementation before my baud bucket fix could end up with the baud value being calculated to be a negative value, which wrapped around, and the user wound up with a non-functional result (eg, a MBAUD=254 or something. because it calculated needing -2 )

rneurink:
yeah the style checker is a pain in that there's no easy way to comply with it. 

I'll be checking overflows etc in a bit, do you have a list of all possible CPU_clocks that can be used? That'll make the bounds checking a bit easier to implement

SpenceKonde:
Uh, well the F_CPU options from the tools -> clock speed? Are you another PIO person? Does ANYONE use the arduino IDE anymore?

At one point the stylechecker was run with suggestions automatically taken... A while later, when I was next editing the tinyNeoPixel libraries.... I didn't figure out what had destroyed the formatting (astyle doesn't like inline assembly or things like the gamma tables much) until I committed and the CI complained about it.
When it's just indentation that it's pissed about and it is demanding something less readable than what you have, you can enclose the relevant part of your code in these comments and it will turn off the style check (I think just for indentation) 
```
// *INDENT-OFF*  

// *INDENT-ON*  
```

SpenceKonde:
20/16/10/8/5/4/1 are the big ones, 24 will be relevant in a couple of versions when overclocking via tuned internal oscillator becomes a thing. 25 is a;lso very achievable and will be an option, but I doubt anyone will care about it. The core also offers (currently just for ext. clock, of course)  28 and 32 as menu options; they don't matter so much though. 32 in particular should probably not even be on the menu, because the tinyAVR parts don't really run at 32...

rneurink:
Hah i use vs code and do a lot of baremetal but translate that to the IDE, actually never used PIO 😅. I'll run a small code test to calculate all baud rates and to implement a safety for it.

rneurink:
By making the baud variable an int16_t we use a tiny bit more memory but makes the bound checking really easy.
Basically this suffices:
`if (baud < 1) {
    baud = 1;
  }`

This is however with in mind that the user won't get any errors back. What is your current take on returning errors to the user?
[baudCheck.xlsx](https://github.com/SpenceKonde/megaTinyCore/files/6471152/baudCheck.xlsx)

rneurink:
image


For now the estimation is not that bad at all as you can see from the measurements and calculations. It of course gets a bit harder with the higher frequency as the baud settings are bound by a uint8_t and the difference between the next integer is quite large

SpenceKonde:
Dayum, that's some next-level care and attention to doing exactly what the code is asked for! The norm within Arduino-land is often a bit halfassed on that front, you know? 

I like it.

It looks like the flash overhead of the bounds-checked version is no worse than of the original PR (flash is what we care about).
 We don't care about an extra byte of transient ram use, unless it's in some function that's likely to be getting exercised when a maximum number of local variables are already defined.or something like that.

Error reporting? It's embedded C. And it's Arduino. Have you forgotten we're in Arduino L:and?  The Wire API pretty much precludes any mechanism of reporting errors anyway. So that's all good. 

What's frustrating is how stuff in classes is that it doesn't get optimized like everything else, and you can'd do compiletime error checking like you can with non-member functions (ie, with __builtin_constant_p() - I tried that to check the baud rates passed to Serial; it don't work (as in, I tried to have it error at compile time if ut sees you passing a constant with a value that can;'t be made at the clock speed selected at compile time 

Say, I just realized,ah, how high do you think this could plausibly go before some intermediate overflows and values suddenly get wacky? The reason I asked is that I just remembered I need to replace this same calculation on DxCore as well. which goes up to 32 MHz at least... and as I saod above, I a, optimistic about doing better with an external oscillator. I'm pretty confident that the 48 that I put external clock menus up to ihn n/

rneurink:
as it uses a uint32_t there won't be an issue up to 4000MHz or so. I see the same rise time and calculations are used in the DA series. 
The only thing is I'm not sure what the silicon offset would be like with these higher frequencies so it can be a bit off.

SpenceKonde:
What do you measn by silicon offset?

SpenceKonde:
I mean, there is one bloody obvious problem with these formulas, that being that we are assuming values for tRise rather than usuing actual values. If I am not mistaken, those should be measured from the system once all devcices and stuff are connecrted to it.

rneurink:
The equation given in the datasheet is a 'perfect' situation. In silicon, there are differences between the simulated model and the real world.  So there's a bit of an offset that has been introduced which I have divided in different sections. 
<= 100kHz,  <= 400kHz <= 800kHz <=1MHz.

so far the lowest frequencies so 10-50kHz are a bit off so I could create another offset for that but I have not seen any I2C devices that operate on such a low frequency. I think this is good enough for now as 100kHz 400kHz and 1MHz (most used) are spot on

PR #427, opened 2021-05-13
emabled => enabled

fredolaredo:


SpenceKonde:
THAAANKKSSSS Always appreciate spelling and grammar fixes. The programmers's test editors don't have that, and codespell isn't very good.

PR #489, opened 2021-07-22
update installation.md

dattasaurabh82:
Added detailed instructions for Arduino-cli


SpenceKonde:
Nothing wrong with the PR itself (well, other than it failed codespell and the newly added and infuriatingly pedantic markdown linter), just wanted to reorganize how alternative IDE information was located within the core documentation. If there was a way I could mark it as your contribution, I would, but there doesn't seem to be a viable way to do that.

PR #530, opened 2021-09-24
Found a bug in TWI_MasterRead

MX682X:
plus I managed to run the cpplinter, so code is a bit more linted

### DIFF
--- a/megaavr/libraries/Wire/src/twi.c
+++ b/megaavr/libraries/Wire/src/twi.c
@@ -675,7 +675,7 @@ void TWI_HandleSlaveIRQ(struct twiData *_data) {
       default:
         // Abort operation
         _data->_module->SCTRLB = TWI_ACKACT_bm | TWI_SCMD_COMPTRANS_gc;
-         while (true) {}  // while loop for debugging. Use this to see what SSTATUS is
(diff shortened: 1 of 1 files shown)


SpenceKonde:
Okay, I'm going to merge this, and then tweak some of your linting. There is no holy book that defines (or would that be declared) what coding styles are to be punished with recursive damnation; unless you strongly object. While there are plenty of coding practices everyone can agree are evil and bad, a significant part of it comes down to taste and consistency with the rest of the core. I have effectively built up a style that I use throughout the cores, (the one that seems most relevant here is `while(condition)`; vs `while(condition) {}`).

On the topic of a timeout (I notice that comment went?).... that is something people have asked for multiple times - We can't use millis/micros or the timers for a timeout (they could be used for something else).... but we can just count passes through the loop that checks for the flag. Obviously this only makes sense when both the bus lines are idle... If the slave sat down on one or both I2C lines and dozed off, that's probably not as useful to time out on, since all further TWI transactions will fail as a result - but the case where the slave simply.... fails to notice the request, or one of the I2C devices failed in a way that doesn't molest the TWI lines., a way to get a timeout would be nice (that is the situation which the users who want that *claim* is occurring in their cases)

PR #568, opened 2021-12-19
Fix ADC and DAC Reference link in README

Modac:
Changed link to point to megaTinyCore analog reference instead of DxCore.


SpenceKonde:
THANKS!

PR #630, opened 2022-01-27
Improve readTempVcc.ino example code

jvasileff:
The code and commit messages should speak for themselves, but I'd be happy to add any detail as necessary.


jvasileff:
It's at least clear that my chips require `-`. I discovered this with two 2-series chips, where one was reporting a sub-freezing core temp, and another finger-burning hot until flipping the sign to match the datasheet, which resulted in sane values. So I then checked a few 1-series chips I have, with the results in the commit message. Maybe older 1-series chips had the sign flipped?

SpenceKonde:
Thanks!

PR #674, opened 2022-03-25
Fixed wrong return type.

7-rate:


SpenceKonde:
Thanks. I'm not sure how those quotes got there - I'm surprised it compiled like that.

PR #702, opened 2022-05-09
Clarify wiring.md

clarkbremer:
I just went through this process, and looked at the docs with fresh eyes.  These are my suggestions.  Make it more obvious how easy it is to wire up, by pulling detailed info out into footnotes.  Tighten up the grammar in spots.


SpenceKonde:
Thank you, this is exactly the kind of thing I need help with! w/regards to docs. The spellchecker keeps adding new words to their misspelled words list (it's actually an inverse spellchecker as far as I can tell, so that it doesn't flag variable names, but does flag obvious typos of real words) .

And we get the trailing space failures because even after years of having this test, I cannot break the habit of putting a space at the end of every line, which was drilled into me when I started learning to type, and which I spent the rest of my life reinforcing the habit until I was told that those extra spaces are evil and terrible and we're going to start calling you Dr. Trailling Whitespace. The only way I can write tests that pass this is if I use an editor that automatically clips them off for me. don't know how to break the habit,

And yeah, I failed to fix a broken link at some point - liks keep getting broken and having to be fixed because of the ongoing docs changes

PR #728, opened 2022-06-20
Changing UART inheritance from HardwareSerial to Stream

MX682X:
since the Arduino API is using forced to be overwritten virtual function (by making them ` = 0; ` all of the API functions are implemented. by skipping this, only the needed functions should be compiled. Effect: significantly reduces codesize

2. The Stream Class is using a static `_startMillis` variable for timedRead and timedPeak. since I can see no use in it except for some debugging applications without an actual debugger, I suggest to deprecate that and save 4 bytes of RAM in EVERY class that uses stream. Also tried to fix the timedRead and timedPeak for anyon who wants to use it without millis, which is noone, otherwise someone would have complained about the fact that the millis define was renamed....

I was able to reduce the binary size on small programs (that use only read or write) significantly.

Anyway, the problem is, I have no time today to check if it works, and actually, not even a tiny at hand. Will test it on an DA tomorrow. Still opened the PR so you know that it is worked on.

### DIFF
--- a/megaavr/cores/megatinycore/HardwareSerial.h
+++ b/megaavr/cores/megatinycore/HardwareSerial.h
@@ -5,5 +5,6 @@
 #ifndef HWSERIALNAME_HACK_H
   #define HWSERIALNAME_HACK_H
   #include <UART.h>
-  #define HardwareSerial UartClass
+  class HardwareSerial : public UartClass {
+  }
 #endif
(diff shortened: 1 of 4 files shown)


SpenceKonde:
I like this change, as I said earlier, I'm going to take it a little further to actually get rid of the unused variable and correct the assembly

PR #730, opened 2022-06-29
Full first run through README.md to tidy up.

SimonMerrett:
Hi Spence, 

I saw in a discussion that you'd be keen for tidying up in the documentation. So I've started with README.md and will send PRs as and when I do a session. Let me know if this isn't welcome. If you have any style guidance, I can try and apply it. 

Simon


SpenceKonde:
Wow! This is beautiful. You are a hero.

SimonMerrett:
I'll keep going as and when I can. What's the best way to ask you about unfinished sentences etc? At the moment my approach has been to trim stuff out if I can't work out the intention. But I'm happy to discuss. I also have to go back round and implement some title case and code quotes from the beginning.

Longer term, if we retain sufficient energy, it may be good to start transferring over to a github wiki, which looks like it supports more structure.

SimonMerrett:
Ps, I keep getting notifications that the PRs are failing CI checks. Is there anything I can do to make them pass? Or is it nothing to do with docs?

SimonMerrett:
I've fetched latest, and finished the first run through README.md. It will need another going over once you / we choose some conventions but is a good first pass. Let me know what do you would like tidying next.

SimonMerrett:
@SpenceKonde I'm not sure if these CI check fails (one of which looks to be in Ref_timers, rather than stuff I've touched) is the reason you've not merged yet, or whether you just need more time to look at the diffs. If it is the CI test failures, I can get working on passing those as a priority.

PR #742, opened 2022-07-16
typos in Ref_Analog.md

ObviousInRetrospect:
Fix two typos.  One of them was in incomplete thought that I didn't know how to complete so I removed it: "You probab"<eol>. If you remember the thought...

The other might have been intended as a demonstration that numbers over the spec are treated as the spec but the choice of 3000 with the comment then using 2000 multiple times seemed more like a typo. If it was intended to be illustrative, perhaps:

Serial.println(analogClockSpeed(2345)); // sets prescaler such that frequency of ADC clock is as
// close to but not more than min(2000, 2345) (kHz) where 2000 is the maximum supported according to the datasheet.
// Any number of 2000 or higher will get same results.


SpenceKonde:
Thanks!

Issue #716, opened 2022-05-26
Linux Upload Fails


dattasaurabh82:
## System: 
**OS:** Ubuntu 22.04 LTS x86_64
**Kernel:** 5.17.8-051708-generic
**IDE:**  1.8.19 (This IDE version works on mac and win)

## Issue:
Upload failure. 
Guessing yaml lib is the culprit. 

## Arduino IDE verbose: 
```shell
Arduino: 1.8.19 (Linux), Board: "ATtiny3227/3217/1627/1617/1607/827/817/807/427/417, ATtiny1607, 5 MHz internal, 1.8V (5 MHz or less), Disabled/Disabled, EEPROM retained, Enabled (default timer), UPDI (no reset pin), 0ms, Master or Slave (saves flash and RAM), Default (doesn't print floats, 1.4k flash use), On all pins, new implementation."

[...]

PR #739, opened 2022-07-12
Updated PyYAML to PyYAML-6.0

InstantMuffin:
This fixes https://github.com/SpenceKonde/megaTinyCore/issues/716

PR #743, opened 2022-07-16
Fixed Hashable Error

dork3nergy:
I was getting an error "collections not hashable" when trying to use this library.  I fixed it by adding a version check for python and then using the correct command syntax based on that version.  It's a decent workaround for now.


InstantMuffin:
@SpenceKonde, favor this over #739 to fix #716.

SpenceKonde:
Thanks - now let's see if it works on windows!

dork3nergy:
Cool.  I'm just glad I could help. I've got my fingers crossed that will
work on windows. Not that that will help.

PR #759, opened 2022-08-06
fix incorrect clock prescalers for 4mhz and 2mhz - new version, does not touch boards.txt

ObviousInRetrospect:
this is a clean diff to just fix wiring.c prescalers had comments and code that didn't agree and the 2mhz was still 1mhz.


SpenceKonde:
Agree, that's the same thing I was going to do about it

PR #775, opened 2022-08-24
possible bug fixes for Serial and TWI

timshac:
Thank you for your excellent software.

I am using an ATTiny3224.

I was having trouble with Serial print flushing from an ISR, and using TWI when also trying to sleep. These changes helped me. I'm not positive they are are correct for all situations.  Please see what you think.

Thank you!

### DIFF
--- a/megaavr/cores/megatinycore/UART.cpp
+++ b/megaavr/cores/megatinycore/UART.cpp
@@ -364,7 +364,7 @@
         // -Spence 10/23/20
         // Invoke interrupt handler only if conditions data register is empty
         if ((*_hwserial_module).STATUS & USART_DREIF_bm) {
-          if (_tx_buffer_head != _tx_buffer_tail) {
+          if (_tx_buffer_head == _tx_buffer_tail) {
             // Buffer empty, so disable "data register empty" interrupt
             (*_hwserial_module).CTRLA &= (~USART_DREIE_bm);
 
--- a/megaavr/libraries/Wire/src/twi.c
+++ b/megaavr/libraries/Wire/src/twi.c
@@ -529,7 +529,7 @@ void TWI_HandleSlaveIRQ(struct twiData *_data) {
   #endif
 
   uint8_t clientStatus = _data->_module->SSTATUS;
-
+   _data->_module->SSTATUS &= ~(TWI_APIF_bm | TWI_DIF_bm); // remove Status flags to prevent ISR from being called again
   #if defined(TWI_MANDS)
     _data->_bools._toggleStreamFn = 0x01;             // reroute stream functions to slave elements
   #endif


ObviousInRetrospect:
@timshac please take a look at https://github.com/SpenceKonde/DxCore/issues/322 - dxcore and mTC share a wire library and there are significant problems with sleep in the current version if you are using it as a slave. Everything in that issue applies to mTC as well.

the author of the library provided a fixed version that I have tested tens of millions of operations against in sleep_mode_pwr_down (and plenty in standby) - previously corruption and hangs were frequent when sleeping

ObviousInRetrospect:
@SpenceKonde I also have sleep related Serial bugs that I haven't characterized well enough to report but the serial fix here looks like it might be related. (Essentially serial wake == data corruption in the rx direction reliably outside of sleep mode idle).

@timshac in general don't Serial print and certainly don't flush from an ISR. I am not even sure that can work, if you are in your ISR that might block the ISRs needed to complete the flush? ISRs should be fast. Instead write a buffer and flag and have the main loop do the slow stuff. There are always exceptions but I have run into problems literally every time I thought I could do Serial things in an ISR.

```
volatile char msg[128];
...ISR...{
    sprintf(msg,...);
 
}

void loop(){
...
  if(msg[0]){ 
      Serial.print(msg); 
      msg[0]='\0';
   }
}

timshac:
> 
> ```
> volatile char msg[128];
> ...ISR...{
>     sprintf(msg,...);
>  
> }
> 
> void loop(){
> ...
>   if(msg[0]){ 
>       Serial.print(msg); 
>       msg[0]='\0';
>    }
> }
> ```
Thanks this is a nice way.

SpenceKonde:
Are you trying to wake on start of frame detection? If so review th errata. Basically in the ISR you *must* turn it off and only turn it on before going to sleep. - because when the data register is read, the next high to low transition will be treated as the start ot the next character - even though it'[s likly in the middle of a character by then.  They really screwed the pooch on that feature. I think the bug is present on all modern avr, even though it's not in all the errata sheets.

SpenceKonde:
Oh - and if you are trying to wake on usart start of frame detection, that isn't expected to work - I just saw where I have incomplete code to make that work ifdef'ed out. Microchip's errata there is a bitch, and even they don't seem to know which parts are impacted because I've seen it appear and disappear from errata lists.....

timshac:
With Serial, I really liked how you wrote Serial.flush() so that it works even when interrupts are disabled, by calling _poll_tx_data_empty()

My proposed change is to disable the DRE interrupt only when the TX buffer is empty, and not before, as the polling loop in flush() exits when the DRE interrupt is disabled.
in _poll_tx_data_empty():
`  
      if (_tx_buffer_head != _tx_buffer_tail) {  // disable "data register empty" interrupt
`
becomes (in my PR)
`
      if (_tx_buffer_head == _tx_buffer_tail) {  // disable "data register empty" interrupt
`

Getting USART start of frame detection working would also be great - I'm sorry to hear it isn't possible probably :(

SpenceKonde:
> 



> With Serial, I really liked how you wrote Serial.flush() so that it works even when interrupts are disabled, by calling _poll_tx_data_empty()
> 
> My proposed change is to disable the DRE interrupt only when the TX buffer is empty, and not before, as the polling loop in flush() exits when the DRE interrupt is disabled. in _poll_tx_data_empty(): ` if (_tx_buffer_head != _tx_buffer_tail) { // disable "data register empty" interrupt` becomes (in my PR) `if (_tx_buffer_head == _tx_buffer_tail) { // disable "data register empty" interrupt`
> 
> Getting USART start of frame detection working would also be great - I'm sorry to hear it isn't possible probably :(

Huh, I wonder why that stupid bug in _poll_tx_data_empty never came up before, it was definitely wrong. 

SFD is implemented. You must not enable the SFD interrupt itself (RXSIE must be 0), putting us on line 3 of table 27-6, and we have to disable it whenever we receive a character, so you must reenable it every time you go to sleep if you want to wake on USART.

PR #793, opened 2022-09-09
Ignore RTC CMP interrupt in millis()

jvasileff:
By ignoring CMP interrupts and only incrementing timer_overflow_count on OVF interrupts, user code may use the CMP interrupt to wake the microcontroller even when using MILLIS_USE_TIMERRTC

---------

With this PR, user code can trivially set `RTC.CMP` to wake up from sleep while still tracking `millis()` using the RTC. Example code below:

```cpp
void setup() {
    // use standby sleep (not power down) since we need the RTC
    set_sleep_mode(SLEEP_MODE_STANDBY);
    sleep_enable();
    RTC.INTCTRL &= ~0x2; // disable CMP interrupt (for good measure)
    RTC.INTFLAGS = RTC_CMP_bm; // clear CMP interrupt (INTFLAGS == 2 at poweron???)
}

void loop() {
    ...
}

void sleep(uint32_t forMillis) {
    uint32_t startMillis = millis();
    uint32_t endMillis = startMillis + forMillis;

    for (uint32_t currentMillis = millis();
            (currentMillis - startMillis) < forMillis;
            currentMillis = millis()) {
        sleepForUpToMillis(endMillis - currentMillis);
    }
}

void sleepForUpToMillis(uint32_t targetMillis) {
    if (targetMillis <= 2) {
        // avoid inadvertantly sleeping for a very long time if there is a delay longer
        // than targetMillis between setting RTC.CMP and calling sleep_cpu() for very
        // short targetMillis (we'll assume no delay will last more than 2ms, set higher
        // if there are very slow ISRs)
        return;
    }

    uint16_t ticks;

    if (targetMillis >= 64000) {
        // limit sleep to less than rtc period
        ticks = 65535;
    } else {
        // each tick is 1000*32/2^15 ms, simplified 1000/1024 ms
        ticks = targetMillis * 1.024; // div 1000 mul 1024
    }

    RTC.CMP = RTC.CNT + ticks;
    RTC.INTCTRL |= 0x2;
    while (RTC.STATUS && RTC_CMPBUSY_bm); // wait for RTC.CMP sync, may not be necessary
    sleep_cpu();
    RTC.INTCTRL &= ~0x2;
    while (RTC.STATUS & RTC_CNTBUSY_bm); // avoid stale millis(), definitely necessary
}
```

### DIFF
--- a/megaavr/cores/megatinycore/wiring.c
+++ b/megaavr/cores/megatinycore/wiring.c
@@ -171,7 +171,13 @@ inline unsigned long microsecondsToMillisClockCycles(unsigned long microseconds)
       timer_millis += 2;
     #endif
   #else
-    #if !defined(MILLIS_USE_TIMERRTC) // TCA0 or TCD0
+    #if defined(MILLIS_USE_TIMERRTC)
+      // if RTC is used as timer, we only increment the overflow count
+      // Overflow count isn't used for TCB's
+      if (RTC.INTFLAGS & RTC_OVF_bm) {
+        timer_overflow_count++;
+      }
+    #else // TCA0 or TCD0
       uint32_t m = timer_millis;
       uint16_t f = timer_fract;
       m += MILLIS_INC;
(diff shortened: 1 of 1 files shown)


SpenceKonde:
Has this change been tested on parts both with and without the brutal RTC errata present on most 0/1-series parts (but not the 3216 or 3217 except in very old specimens - most are now Rev. C which fixed it, no other parts have gotten such a fix though). If not this isn't going into todays release

jvasileff:
I developed on a `1617` and briefly tested on a `3226`.

Can `SYSCFG.REVID` be used to reliably determine the rev? On both chips, `Serial.println(SYSCFG.REVID);` reports `0`.

SpenceKonde:
Only one version of the 1617 has been released. Only one version of the 3226 has been released. All 1617 have the bug, all 3226 do not have the bug.

PR #804, opened 2022-09-14
add missing #endifs

ObviousInRetrospect:
this now compiles.

still overall hitting different UART issues.

Also personally had to get rid of -Wall to figure anything out as a a build of a simple sketch that used none of the constants generated 6000 + lines of warnings about constants. Will figure out how to turn that message off but the core itself generates those warnings.

### DIFF
--- a/megaavr/cores/megatinycore/core_devices.h
+++ b/megaavr/cores/megatinycore/core_devices.h
@@ -3735,91 +3735,113 @@
     #define USART_RXPL_3_bm USART_RXPL3_bm
   #elif defined(USART_RXPL_3_bm)
     deprecated_constant_name USART_RXPL3_bm = USART_RXPL_3_bm;
+  #endif
   #if !defined(USART_RXPL_3_bp) && defined(USART_RXPL3_bp)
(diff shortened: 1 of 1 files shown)


SpenceKonde:
Thanks. And hell, is there no way to make a warning that only shows when you use the bloody constant!?!!

I would kill for a way to make a warning show up based on compiletime known invalid values.   `void badArg(const char*) __attribute__((error("")));` errors if the function is referenced in theoretically reachable code - there's an analogous warn() or warning(), but nobody I've talked to has managed to make it work. okay, no, maybe not *kill* for, but if I knew who had the knowledge and where they live, well, if they refused to tell me, they'd wish I was (willing to kill for it, that is.... 

I doubt they'd refuse though - you'd be amazed at how responsive support personnel are when you make a "site visit" at their home....

ObviousInRetrospect:
What if you guarded the obnoxious defines with an ifndef something that you then define in all your c/c++ before including that header.

Or had two versions one without the obnoxious stuff with the same multiple inclusion guard define/ifdef and explicitly pull in the quiet one at the top of all your files.

PR #807, opened 2022-09-15
more core_devices.h fixes

ObviousInRetrospect:
guarded the warnings with #ifdef ENABLE_DEPRECATED_MICROCHIP_WARNINGS as a prelude to a solution to not trigger them from mTC code.

added a missing = sign

still hitting errors I don't understand in 2.0 (now released, no longer RC):
```
/Users/user/Documents/GitHub/megaTinyCore/megaavr/cores/megatinycore/core_devices.h:1931:30: error: expected unqualified-id before numeric constant
     deprecated_constant_name PORT_INT0_bm = PORT_INT_0_bm;
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 megaavr/tools/python3/*
 *.pyc
 *.pyc
+.DS_Store
--- a/megaavr/cores/megatinycore/UART0.cpp
+++ b/megaavr/cores/megatinycore/UART0.cpp
@@ -82,7 +82,7 @@
 #else
               "rjmp   _do_rxc"    "\n\t"
 #endif
-              ::"z"(&Serial1));
+              ::"z"(&Serial0));
         __builtin_unreachable();
     }
   #endif
--- a/megaavr/cores/megatinycore/core_devices.h
+++ b/megaavr/cores/megatinycore/core_devices.h
@@ -1925,9 +1925,14 @@
   #endif
 
   /* ======= PORT ======= */
-  #if !defined(PORT_INT_0_bm) && defined(PORT_INT0_bm)
+
+/* commenting out for now - not sure what if anything should replace this but was hitting the #error I inserted because these are defined and being set as constants
+#if !defined(PORT_INT_0_bm) && defined(PORT_INT0_bm)
     #define PORT_INT_0_bm PORT_INT0_bm
   #elif defined(PORT_INT_0_bm)
+    #if defined(PORT_INT0_bm)
+      #error("FUBAR - will cause compile error below")
+    #endif
     deprecated_constant_name PORT_INT0_bm = PORT_INT_0_bm;
   #endif
 
@@ -2021,6 +2026,7 @@
     deprecated_constant_name PORT_INT7_bp = PORT_INT_7_bp;
   #endif
 
+ */
   #if !defined(PORT_ISC_0_bm) && defined(PORT_ISC0_bm)
     #define PORT_ISC_0_bm PORT_ISC0_bm
   #elif defined(PORT_ISC_0_bm)
--- a/megaavr/cores/megatinycore/core_devices.h
+++ b/megaavr/cores/megatinycore/core_devices.h
@@ -518,13 +518,17 @@
   // So they went ahead and made that change. That is what's called a "breaking change", really for no reason except codes style. Most companies even if they decided to go that route, would never do that without introducuing a compatibility layer.
   // That wanton disregard for backwards compatibility is not acceptable in an Arduino core nor in a commercial product.
   // Using the old names will produce warnings. These deprecated names should be fixed as support for these FOUR THOUSAND LINES of bandaids WILL BE REMOBVED in 1.6.0!
+#ifdef ENABLE_DEPRECATED_MICROCHIP_WARNINGS
   typedef const uint8_t __attribute__ ((deprecated("\nMicrochip changed the spelling of bits within a bitfiels (macros that end in the bitnumber followed by _bm or _bp), you are using the old name, ex PERIPH_BITFIRLD1_bm.\nYou should use PERIPH_BITFIELD_1_bm; we do not guarantee that this 4000-line bandaid will not be removed in the future.\r\nWhy did they do this? Beats me. Ask their support folks - if enough of us do it, they might hesitate next time they have the urge to mass rename things in their headers")))  deprecated_constant_name;
+#else
+    typedef const uint8_t deprecated_constant_name;
+#endif
 
   /* ======= ACs ======= */
   #if !defined(AC_HYSMODE_0_bm) && defined(AC_HYSMODE0_bm)
     #define AC_HYSMODE_0_bm AC_HYSMODE0_bm
   #elif defined(AC_HYSMODE_0_bm)
-    deprecated_constant_name AC_HYSMODE0_bm AC_HYSMODE_0_bm;
+    deprecated_constant_name AC_HYSMODE0_bm = AC_HYSMODE_0_bm;
   #endif
 
   #if !defined(AC_HYSMODE_0_bp) && defined(AC_HYSMODE0_bp)


SpenceKonde:
That's the right assembly change. 

my plan for pins_arduino is a bit different though. Since there doesn't actually seem to be a way to make a deprecated constant that only gives a warning when used, I'm just going to regex them into #defines.

SpenceKonde:
I've addressed the core_devices.h issue slightly differently. 

Still getting a strange linker error: 
c:/arduino-1.8.13_azd5test/hardware/tools/avr/bin/../lib/gcc/avr/7.3.0/../../../../avr/bin/ld.exe: address 0x80380a of C:\Users\Spence\AppData\Local\Temp\arduino_build_261672/sketch_sep14a.ino.elf section `.bss' is not within region `data'

ObviousInRetrospect:
>Since there doesn't actually seem to be a way to make a deprecated constant that only gives a warning when used, I'm just going to regex them into #defines.
So I actually think you can get conditional warnings if you want them and started down that path with my version.

You #define something like DMT_CORE_INTERNAL at the top of every .c and .cpp file in the core that uses deprecated constants (why not just use sed to fix the constants instead, do you want me to do that and submit a PR that changes all the old constants to new constants)?

You use the conditional define to have your compilation units treat them as uint8_ts and the user's compilation units treat them as deprecated uint8_ts. 

I do wonder whether #defines vs constants affect the assembly. I suspect not but would verify.

SpenceKonde:
How does that not result in the warning showing up if the constant isn't used, though? What separates the that from the case where we define deprecated uint8_t's for both core and user files? The core has like 1 reference to a deprecated constant, total. The problem appears to be that simply declaring the deprecated type in a header and giving it a value is considered referencing it, and hence reason to give a warning.

Why not sed? because my development system is windows, and because I know regexes and not sed. I already use regexes very frequently - they are an excellent "hammer" and are particularly valuable in the type of tasks that core development often takes (like, say, copying every constant with a name ending in an _ followed by a digit followed by _bp or _bm into another file - those are the new names. There were few if any cases that had to be manually removed. Sublime Text lets you "find all" and copy-paste to another file. Then you can use those as inputs to the next regex and use actual regex replacement there to slice, duplicate and dress the inputs as required. Once you're comfortable using regexes, they are very good at writing code that would otherwise be difficult to automate creation of. It doesn't get you everything (because you can't do math in a replacement expression, but it is good at boiling inputs down to single lines and separating the numbers from words, and putting commas between fields, so when I need to do that, I can just generate a .csv file, open in excel - which is great at repetitive math, but struggles with text - and make columns for the new data, copy the column and paste special -> values over the old data and delete the temp column as needed. Often I find that I don't even have to save the intermediate in excel, I can just directly paste into sublime and work with that. Either way, once it's back in sublime, it's a simple matter to reconstitute the required code. 

For inline assembly, #defines are treated like find replace as long as they're not within quotation marks - So defines are actually much more powerful, though at times they look like something from the lunatic fringe. This is from a Flash.cpp library included with DxCore, where depending on the configuration, we either know the absolute address of an spm Z+ follower by ret - or we have the name of a label that has those two instructions - or we restrict writes to APPDATA section and can do the spm directly.
0x1FA is 3 words before the end of the bootloader section, where I stash those two instructions immediately before the optiboot version number). Similarly you can put #ifdef's the middle of assembly - see UART.cpp. The preprocessor doesn't know the difference between inline asm and C - it's a text processor. 
```
#if defined(USING_OPTIBOOT)
  #define SPMCOMMAND "call 0x1FA" 
#elif defined(SPM_FROM_APP)
  #if SPM_FROM_APP == -1
    #define SPMCOMMAND "call EntryPointSPM"
  #else
    #define SPMCOMMAND "spm Z+"
  #endif
#else
  #error "You must also enable writing to flash from app in tools menu."
#endif
// bunch of code, but when it comes time to write the flash:
  __asm__ __volatile__(
            "movw r0,%A[dat]"                     "\n\t"
            SPMCOMMAND                            "\n\t"
            "clr  r1"                             "\n\t"
            : [flptr]   "+z"   (zaddress)
            : [dat]      "r"   (data)
          );
```

As far as the preprocessor's treatment, it's treated like any other function. That the compiler treats the functions very differently - simply concatenating all the strings before the first : then picking registers than meet your constraints, so it can do the operand replacement, initialization, and cleanup if necessary (the cleanup happens if it it had to move a local variable that you later reference out of the way to meet your constraints when picking registers for operands, especially the powerful Y and Z pointer registers, which can do load/store with *displacement*, and since the compiler frequently has Y set to the stack pointer in order to access local variables on the stack. 
**below here I veer way off topic**
You're always wishing that you could do displacement on the X pointer (which would require 4096 more opcodes out of the 1500 unused 16-bit opcodes, making it a clear impossibility), or that there was a W pointer (ie, r24:r25, which they totally had the unused opcodes for, and which is already a "special" register pair like X, Y and Z because you could use ADIW and SBIW on it. Maybe if there were two of them, they'd have put more care into compiler treatment than they did for the single X pointer, too, which would be nice. When it runs out of pointer registers that can do displacement, it's next resort is adiw, ld/st X, sbiw - they implement it as a virtual ldd/std X+q in one layer, which only later in the compilation process gets translated into real instructions. Such accesses frequently occur consecutively in dense clusters, and *immediately consecutive adiw and sbiw are not combined*, making a store take 5 clocks and 3 words instead of 1 clock / 1 word, and a load take 6/3 instead of 2/1 (every time two of them appear consecutively, combining the adiw and sbiw would save 2 / 1, but as I said, it misses even that obvious optimization.  

A thicket of those was part of why the serial ISRs were reimplemented, because in that context the speed really is critical, particularly since modern AVRs can actually reach the maximum baud rate, which is close to the maximum rate that characters can be transferred between the software ring buffers and the hardware TXDATA/RXDATA registers. I think the ISRs are currently slightly slower than the hardware serial, and thus every extra clock cycle lowers the maximum sustained baud rate. At 16 MHz that's 2 mbaud hardware (80 clocks per character for 8 bit  characters with 1 start and 1 stop bit, and 200 bytes/ms. So you can receive up to RX_BUFFER_SIZE characters reliably at max baud rate only if (RX_BUFFER_SIZE - 2) * (ISR exec time + isr overhead)+(millis isr exec time incl. overhead) < 80 * RX_BUFFER_SIZE. Since every working register used in an ISR has to be saved and restored (adding 3 clocks apiece), and the isr overhead is 9-10 clocks absolute minimum depending on flash size if neither working registers nor the SREG are changed in the ISR, and that increases to 20-21 unless you write naked isrs completely in asm), 80 clocks is a very ambitious goal, especially when you have to fish offsets out of the HardwareSerial class member to find the ring buffers. And we also have to do some other stuff for various reasons to support some features (wake on SFD requires 4 clocks to work around the universally present errata, and keeping a record of errors reported by the hardware is required for autobaud and desirable generally - that takes another 6 or so - both in the RXC interrupt. All told I just calculated that the RXC ISR is 79 clocks on parts with 8k or less flash, up to 81 otherwise). So we squeak by that metric for RXC - though ofc if we need to receive a continuous stream of data longer than that, it has to spend enough time between ISRs for the user code to grab the data out of the ring buffer, and we don't control the user code, but even really tight code is going to knock the true continuous maximum speed to probably half the hardware maximum. But anyone expecting to receive arbitrary lengths of serial data at the maximum hardware baud rate doesn't have reasonable expectations :-) . An application specific implementation of Serial would help, especially if you knew when data was going to come (for example through a modem liaison pin, permitting a polling loop instead of an ISR, which would make continuous RX and processing possible at the maximum speed for very simple and specialized applications only, where you don't need to do much processing on the received data - imagine a Serial to Parallel converter (each received byte is written to VPORTx.OUT), or serial to SPI converter at F_SCL >= F_CPU/8 - that sort of thing). And you'd have to use no references to the serial classes to do this (easier than it sounds). 

HyperUPDI is going to do something much like that actually, though I'm targeting less than 1/3rd of the abs. max hardware speed. Most of the SRAM will be one giant ring buffer, and while in data mode, a single loop will be polling for received data and DRE on the half duplex UPDI serial port. Would base it on an AVR64DD20 or AVR32DD20, or a AVR64DD32 for the improved version, which will add an external SPI flash chip to store a more flash images than anyone needs; a 1GBit (128 MByte)  SPI flash IC only costs like 4 bucks, and can be run at F_SCL = F_CPU/2) and very basic UI, allowing standalone programming, which a lot of people want). Would get a new upload tool written in python, because the UPDI protocol would be 100% handled on the programmer, so we could send huge blocks and avoid all the USB latency, thus allowing programing to occur at the the absolute maximum speed that the target can take, and using MVIO, we could program parts at 5v, 3.3v with the programmer supplying power, or let the target supply VDDIO2 (so you could, say, program a part that's powered by a LiPo battery without having to disconnect the battery). It'll be sick if I ever get a chance to do it!

ObviousInRetrospect:
delta between regexes and the parts of sed I use is basically zero. sed -e 's/regex/replacement/g' with capture being \( ... \) instead of bare parens and the capture groups being written as \1 ... \n in the replacement. The / is a delimiter of your choice, conventionally / although | is often used when either the regex or the replacement uses / to avoid escaping. Sure you can do more advanced stuff in sed but it gets unreadable quickly. meta-sed on the other hand can be fun if perverse (using regexes to generate regexes and then applying those second order regexes) and comes in handy more often than you might expect. 

Subsystem for linux 2.0 on windows is actually awesome. My second favorite windows feature behind hyper-v having a decent RDP server. The integration is extremely smooth. pop open a real shell, access to the windows filesystem, do the things that suck less in a linux environment in place.

https://ubuntu.com/tutorials/install-ubuntu-on-wsl2-on-windows-11-with-gui-support#1-overview (which links the w10 version if you prefer)

Ironically I run a windows box that is effectively headless mostly to host VMs because HyperV+RDP sucks less than the VNC crap on the linux side.  I just do most of my development on a Mac laptop instead of in a remote VM because forwarding USB ports to VMs is annoying.

Issue #801, opened 2022-09-14
2.6 comparator library


ObviousInRetrospect:
Got this on a sketch that included had a gratuitous #include <Comparator.h>
```
In file included from /Users/user/Documents/Arduino/soilmoist3224b/soilmoist3224b.ino:19:0:
/Users/user/Library/Arduino15/packages/megaTinyCore/hardware/megaavr/2.6.0/libraries/Comparator/src/Comparator.h:323:5: error: 'in_comparator' does not name a type; did you mean 'Comparator'?
     in_comparator::p::inputP_t     input_p        = in_comparator::p::in0;
     ^~~~~~~~~~~~~
     Comparator
/Users/user/Library/Arduino15/packages/megaTinyCore/hardware/megaavr/2.6.0/libraries/Comparator/src/Comparator.h:324:5: error: 'in_comparator' does not name a type; did you mean 'Comparator'?
     in_comparator::n::inputN_t     input_n        = in_comparator::n::in0;
     ^~~~~~~~~~~~~
     Comparator

exit status 1
Error compiling for board ATtiny3224/1624/824/424 w/Optiboot.
```

when I commented that out I got the USART error again (on a sketch that does use Serial)

ObviousInRetrospect:
I think #809 fixes this

SpenceKonde:
Thannnnkk youuuuu

PR #809, opened 2022-09-15
fix comparator namespace reference

ObviousInRetrospect:
it builds, ship it! (double check that the namespace I pointed it at makes sense)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 megaavr/tools/python3/*
 *.pyc
 *.pyc
+.DS_Store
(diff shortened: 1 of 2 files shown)


ObviousInRetrospect:
fix for #801

SpenceKonde:
Thank you. I wasn't particularly looking forward to this one, as I don't really know C++ very well; I am more comfortable in asm than C++. 

What can I say, I'm just not classy enough for C++ ;-)

ObviousInRetrospect:
Yeah I kinda figured that. I am not a fan of C++ either but I used to know it well (pre C++11) but don't use it often enough to remember all its plentiful warts. Not as bad as perl but still a pretty terrible language. I find the fewer of its features you use the better the results.

PR #815, opened 2022-09-16
fix the new millis asm for 2MHz (tested, was wrong, now correct). Boards.txt :-(

ObviousInRetrospect:
… a missing 2MHz on 3224

- the new millis() asm didn't look right for 2mhz. Went to test it, found the menu item wasn't present on the 3224 I was testing on. Fixed that. Will come back and do a pass making sure it is present everywhere.

After testing determined the conditional change was wrong and fixed it.

testing: before, millis incremented at half speed, delay(1000) in loop behaved strangely (alternate calls took longer) at 2mhz, normal at 4mhz. after: 2mhz and 4mhz behave similarly.

### DIFF
--- a/megaavr/boards.txt
+++ b/megaavr/boards.txt
@@ -10,7 +10,7 @@
 #________________________________________________________________________#
 #                                                                        #
 #  Copyright Spence Konde and other         ##        ###         #      #
-#  contributors 2018-2021                  #  #      #           ##      #
+#  contributors 2018-2022                  #  #      #           ##      #
 #  megaTinyCore is free software.    # #     #       ####         #      #
 #  (LGPL v2.1) see LICENSE.md and    # #    #    ##  #   #  ##    #      #
 #  License section of README.md       #    ####  ##   ###   ##   ###     #
@@ -19,7 +19,7 @@
 #  Now reformatted for easy navigation with text editing tools which     #
 #  have the zoomed out "minimap" that you can scroll through.            #
 #                                                                        #
-#  megaTinyCore 2.6.0-1 is being developed simultaneously with DxCore    #
+#  megaTinyCore 2.6.1 is being developed simultaneously with DxCore    #
 #  1.5.0. As the hardware is very similar, much code is shared, and that #
 #  is the version of DxCore most similar to this version of megaTinyCore #
 #                                                                        #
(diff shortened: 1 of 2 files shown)


SpenceKonde:
We're fixing the same things simuxtaneously x_x

SpenceKonde:
Though the 2mhz option was actually missing everywhere

SpenceKonde:
And the version of asm millis that's going in is significantly different from what it currently checked in, but I'm testing it before I put it in because I don't want any surprises.

SpenceKonde:
2 MHz is still mondo fucked.

SpenceKonde:
Okay, there....

ObviousInRetrospect:
you still probably want this tool

ObviousInRetrospect:
merged: took your version of the ASM, boards changes overlapped

ObviousInRetrospect:
there were a couple other errors in boards.txt that I fixed here (the 412w related stuff had some copy/pastes that missed the edit and some duplicate definitions).

let me know if you need better docs on using the tool I added. it is super quick and dirty but was able to correctly do the clock insertions. It errs on the conservative side and shouldn't introduce problems. Output goes to ../../boards.out which I manually diff before accepting.

SpenceKonde:
Also I'm sorry I haven't had time to really dig into the new pinout image thing much - I'm starting to take some flak for not having DxCore out with the DD support (and megaTinyCore was much closer to working and used a lot of the same code). Since the pinout images aren't distributed with the core, and can be added into the docs after when people started turning up the heat in the past few weeks, I've had to focus on getting working releases out.

ObviousInRetrospect:
I looked at the tool you had and started there but concluded it wasn't done and I wasn't sure how to fix it. Full generation seems better but last time I did hand edits I concluded a crappy tool was less work than hand edits. On your tool I might pull some/all the dict literals out into json (which you can do with json.dump(dict, file,sort_keys=True,indent=4)). I don't know if you use black but its sort of the standard format enforcer. pip install black. python -mblack file.

100% in agreement on 1.5.0 being the priority - I'm far more excited about DD support too.

No problem as long as you don't need the pinout images to go out with the core. There are some questions pending that I sort of want to lock down direction on before I do further work on it as right now I have too many options in the code which makes it hard to work on. In particular if we conclude the new style yellow boxes with PIN numbers are good I can delete a lot of horrible and unmaintainable code which unblocks me from doing some scaling work that is painful if it has to support the numbers in the box. I think the current images are usable as long as I didn't f*ck up the PWM (if I did, fixing it should be fast).

PR #816, opened 2022-09-16
Improve readability of millis IRQ

MX682X:
checked for correct addresses by looking at the assembly. Seems to work. What do you think?

(I'm sorry for "erasing" all the tedious work)

### DIFF
--- a/megaavr/cores/megatinycore/wiring.c
+++ b/megaavr/cores/megatinycore/wiring.c
@@ -153,59 +153,18 @@ inline unsigned long microsecondsToClockCycles(unsigned long microseconds) {
   #elif defined(MILLIS_USE_TIMERRTC)
     ISR(RTC_CNT_vect)
   #else //otherwise it's a TCB *or* no MILLIS_USE_TIMER macros are defined, including MILLIS_USE_TIMER_NONE
-    #if defined(MEGATINYCORE) // tinyAVRs use the dart board approach to locating the peripherals :-(
-      #if defined(MILLIS_USE_TIMERB0)
-        #if (MEGATINYCORE_SERIES != 2)
-          #define _CLEARFLAG "sts  0x0A46, r24"
-        #else
-          #define _CLEARFLAG "sts  0x0A86, r24"
-        #endif
-        ISR(TCB0_INT_vect, ISR_NAKED)
-      #elif defined(MILLIS_USE_TIMERB1)
-        #if (MEGATINYCORE_SERIES != 2)
-          #define _CLEARFLAG "sts  0x0A56, r24"
-        #else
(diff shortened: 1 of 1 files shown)


SpenceKonde:
Not going into 2.6.1. Asm is pretty opaque no matter what to most people, and code readability is not as big a concern as getting 1.5.0 of DxCore out getting pinout images out, and getting ATTinyCore 2.0.0 out.. Once we have a stable version of megaTinyCore released, I'll merge this in, and the people who install manually can start exercising it. But I can't risk 2.6.1 over this, as it doesn't impact functionality. Yes the living room could use new wallpaper, but there's no electricity, half the windows are broken and the roof is caving in. 

I don't even know where you found out about an "m" constraint; it doesn't seem to be documented. 

2.6.1 is being released within minutes, and then my attention will be focused on CI results here and getting DxCore to work

MX682X:
> I don't even know where you found out about an "m" constraint; it doesn't seem to be documented.

I started digging in the compiler source... Also it was mentioned in stackOverflow.
Actually, it is documented, but not in the outdated avr-gcc asm docs ([https://gcc.gnu.org/onlinedocs/gcc/Simple-Constraints.html]())

jvasileff:
> Asm is pretty opaque no matter what to most people

Yes.

It would be really helpful if the original code could be kept as a comment when things are re-written in assembly. This:

```cpp
    #if (F_CPU > 2000000)
      timer_millis++; // that's all we need to do!
    #else
      timer_millis += 2;
    #endif
```

and [the current code](https://github.com/SpenceKonde/megaTinyCore/blob/0cc2f9cd42d7edbc078e500789ce87772859a04a/megaavr/cores/megatinycore/wiring.c#L214) are **miles** apart in readability.

I've had the need to read through the millis code a few times, and getting hit with blocks of assembly would either slow me down significantly, or possibly make me just give up. The problem isn't just trying to read a dozen lines of assembly, but trying to read perhaps a couple hundred lines of C and hitting a bunch of assembly road blocks along the way.

The comment at the top:

```
    // Okay, so this ISR is naked if it's a TIMERB, but clothed otherwise, and the timerb ISR is implemented fully in asm because the compiler was doing
    // it in a particularly boneheaded way.
```

is certainly helpful and informative for the maintenance of the ISR, but provides no clue as to what the ISR actually does.

ObviousInRetrospect:
Are there places in DxCore 1.5.0 where you want help either testing or tracking down specific bugs (to avoid duplicate work on the same bug)? I'd also need a 2 sentence summary of what you think is unfinished/known broken in head of tree. mTC head of tree seems to be working for me in light testing. Put a couple sketches over to it and nothing is obviously broken.

Issue #817, opened 2022-09-17
Can't chip erase with prog.py


Azmisov:
Not sure exactly how it happened, but I've got an attiny424 that got into a locked state and can't be programmed anymore. I'm trying to do a chip erase with prog.py.
Calling: `python3 prog.py -d attiny424 -a erase -u /dev/ttyUSB0`
And the logs:
```
Version 1.2.3 - Jan 2022
Using serial port /dev/ttyUSB0 at 115200 baud.
Target: attiny424
Action: erase
pymcuprog.nvm - ERROR - Device is locked.
Error:
Failed to enter NVM programming mode
Pinging device...
Device ID mismatch! Stopping.
```
I think `pymcu._start_session` shouldn't have returned `STATUS_SUCCESS`, since the device was locked (if it didn't return success, it would have dropped into the chip erase logic). Digging through pymcuprog included in megatinycore, looks like the culprit is in `Backend.start_session` -> `self.programmer.setup_device` ->  `get_nvm_access_provider` -> `NvmAccessProviderSerial.__init__` -> `self.avr.enter_progmode()`. The final call fails (device doesn't unlock before timeout); but the error is caught. Everything else runs fine unwinding the stack to `Backend.start_session`. Looks like a `PymcuprogDeviceLockedError` should have been raised somewhere, but is not. I think the solution is to raise raise `PymcuprogDeviceLockedError` in the try-except block for that `enter_progmode()` call (see nvmserialupdi.py).

(Note I think the Device ID mismatch error is misleading, and I suspect just shows up because the NVM programmer couldn't unlock the device and couldn't get an ID)

So after adding that change, I get:
```
Version 1.2.3 - Jan 2022
Using serial port /dev/ttyUSB0 at 115200 baud.
Target: attiny424
Action: erase
pymcuprog.serialupdi.application - ERROR - Timeout waiting for device to unlock
pymcuprog.nvm - ERROR - Device is locked.
Error:
Failed to enter NVM programming mode: device is locked
The device is in a locked state and is not accessible; a chip erase is required.
Locked AVR UPDI devices can:
 - be unlocked using command: erase --chip-erase-locked-device
 - write user row values using command: write --user-row-locked-device
Locked state detected, performing chip erase
pymcuprog.nvm - ERROR - Device is locked.
Error:
Failed to enter NVM programming mode
The device is in a locked state and is not accessible; a chip erase is required.
Locked AVR UPDI devices can:
[...]

PR #818, opened 2022-09-17
fix chip-erase for serialupdi

Azmisov:
See issue #817, these small changes seem to fix the failed chip-erase I was experiencing


SpenceKonde:
Hot damn, for reals?

That was not expected,. That was the most threatening of the Three Missing Features, has it fallen already? Although not the hardest to implement, it's absence was the most glaring.

PR #825, opened 2022-09-19
minor readme fixes

ObviousInRetrospect:
but mostly trying to force github actions to run now that 2.6.1 is avail


ObviousInRetrospect:
hmm apparently you can evade pull request checks by not including any files that would trigger them and then adding those files later. Kind of the opposite of what I wanted to do. closing and reopening.

PR #826, opened 2022-09-19
220918 docs & 2.6.0+ -> 2.6.1+

ObviousInRetrospect:
Resubmitting and including a version number bump from 2.6.0 to 2.6.1 in core_parameters.h to trigger the regression tests.


ObviousInRetrospect:
The motivation behind this change was to trigger compile-examples now that 2.6.1 is installable. While I think it is an improvement and fixes a markdown-lint error it is a very low value pull-request and I am largely indifferent to whether it is merged.

PR #829, opened 2022-09-20
Updating Wire library to 2.0.9 to be used in mTC 2.6.2

MX682X:
leaving pop/push sleep in TWI for now, can be moved later to somewhere else to be used by other hardware, as soon as it's proven itself "in the field".
Also, posted to mTC since DxCore still didn't get it's 1.5.0 yet


SpenceKonde:
Thanks 
I can't say I understand evertything that went into this.

PR #832, opened 2022-09-27
add asm for TCAx and TCD0 when used as millis Timer

MX682X:
let's not leave the 1-series people out of the 20-30% millis ISR improvement

### DIFF
--- a/megaavr/cores/megatinycore/wiring.c
+++ b/megaavr/cores/megatinycore/wiring.c
@@ -279,13 +279,14 @@ inline uint32_t microsecondsToClockCycles(const uint32_t microseconds) {
       */
     #else // TCA0 or TCD0, also naked
       __asm__ __volatile__(
-      // ISR prologue (overall 8 words / 8 clocks):
-      "push       r24"            "\n\t" // we use two registers other than the pointer
+      // ISR prologue (overall 9 words / 9 clocks):
+      "push       r24"            "\n\t" // we use three more registers other than the pointer
       "in         r24,     0x3F"  "\n\t" // Need to save SREG too
       "push       r24"            "\n\t" // and push the SREG value
-      "push       r25"            "\n\t" // get a second byte for buffer
-      // timer_fract handling (13 words / 16/18 clocks):
-      "clt"                       "\n\t" // Clear Temp bit in SREG, we'll use it later
+      "push       r25"            "\n\t" // second byte
+      "push       r23"            "\n\t" // third byte
+      // timer_fract handling (13 words / 15/16 clocks):
+      "ldi        r23, %[MIINC]"  "\n\t" // load MILLIS_INC. (part of timer_millis handling)
       "ld         r24,        Z"  "\n\t" // lo8(timingStruct.timer_fract).
       "ldd        r25,      Z+1"  "\n\t" // hi8(timingStruct.timer_fract)
       "subi       r24,%[LFRINC]"  "\n\t" // use (0xFFFF - FRACT_INC) and use the lower and higher byte to add by subtraction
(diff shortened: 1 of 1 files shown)


SpenceKonde:
Ported to DxCore. At thatpoint it was observed that delay all timekeeping is completely fucked on a type A timer. Any ideas? Should I rip out of DxCore 1.5.firstversionIactuallypackage just to make sure we have a working core that can be released?

MX682X:
I'm right now like: how the eff have I made the delay nonworking while it does not rely on millis... And I think I might have found the culprit, but I do not think it was me.

https://github.com/SpenceKonde/DxCore/blob/b36ff404729ff6c5ff8385f9fb1c903269cdf375/megaavr/cores/dxcore/wiring.c#L926


```
void _delay_nonconst(unsigned long ms) {
    while (ms > 0) {
      _delay_ms(1);
    }
  }
```

isn't there a decrement missing?

MX682X:
Didn't had time today to actually test out the update, but I've just looked through the wiring.c history and noticed that the post decrement (except I'm missing something, because how has noone noticed an eternal delay loop) had been there for quite some time, so it is likely not the cause. 
I try to look into it tomorrow with a debugger. hopefully I can figure it out. Worst case scenario: DxCore will get the TimerA ISR in C, aka no changes.

MX682X:
I'm not sure if this is the right thread to point it out, but I like to point out that I was not able to upload anything. I've simply replaced the 1.4.10 files with the 1.5.1 from here and got following from AVR-DUDE

Command (multiline for better readability):
```
C:\Users\USER\AppData\Local\Arduino15\packages\DxCore\tools\avrdude\6.3.0-arduino17or18/bin/avrdude 
-CC:\Users\USER\AppData\Local\Arduino15\packages\DxCore\hardware\megaavr\1.4.10/avrdude.conf -v -pavr128da48 
-ccuriosity_updi -Pusb -Ufuse5:w:0b11001001:m -Ufuse6:w:0x04:m -Ufuse7:w:0x00:m 
-Ufuse8:w:0x00:m""-Uflash:w:C:\Users\USER\AppData\Local\Temp\arduino_build_297836/sketch_dec26a.ino.hex:i 
```
 
with following error:

```
avrdude: reading input file "0x00:m""-Uflash:w:C:\Users\EA\AppData\Local\Temp\arduino_build_297836/sketch_dec26a.ino.hex"
avrdude: can't open input file 0x00:m""-Uflash:w:C:\Users\EA\AppData\Local\Temp\arduino_build_297836/sketch_dec26a.ino.hex: Invalid argument
avrdude: read from file '0x00:m""-Uflash:w:C:\Users\EA\AppData\Local\Temp\arduino_build_297836/sketch_dec26a.ino.hex' failed

```


adding a space between fuse8 and flash argument has resolved the problem. 


Also, about the delay: Some tests have confirmed that the decrement is missing. Furthermore, the TCA based millis delay fails ONLY at 24MHz Clock speeds, otherwise it works fine. At least with non-const delays.

MX682X:
Sooo.. In a nutshell, the problem lies in backtraveling of micros somewhere in the fixed point division. Using the fall through option, it works but not as accurate. P.S. : the Division part has 2x 28MHz and no 16MHz option for TCA, a copy paste error I guess. Also also, there are some defines that check for MILLIS_USE_TIMERA0, but not A1, so overall... I'd recommend to wait with the release for now.....

SpenceKonde:
Ugh wait, why is there any division happening in micros?! We can't have division in micros, it is far too slow.

MX682X:
What I mean was bit-shift division. For every right shift you divide the value by two. 

I do not know how it happend, but backwards travel is the only explanation for ((micros() - start_us) > 1000) to fire about 10 times in a row, without wait time. With trial and error I was able to confirm it and decided to check all other frequencies too.

Also, I wanted to outsmart the compiler and reduce the times it has to do bitshifts, but it was no use. Intrestingly enough, the compiler outsmarted me by first doing right shifts and adding the 16-bit multiplicated value later....

Issue #839, opened 2022-09-30
Maximum Wire BUFFER_LENGTH is 255, NOT 256


specternecter:
It took me hours to figure out what was happening, but I can't seem to figure out how to fix it in the library. I have an I2C device that sends 256 byte packets to a slave. I'm using an ATTINY3226 as the slave and I modified the boards.txt file to allow me to select a 256 byte BUFFER_LENGTH. However, Wire.onReceive() never returned any data. So I got out an ESP32 and started sending packets to find out why I wasn't getting anything. It turns out that Wire.onReceive() only works if you send **LESS** than 256 bytes. You can send 3 bytes or 55 bytes, or 255 bytes but you can not send 256 bytes. If you send 256 bytes, Wire.onReceive() seems to overflow somewhere and the function will never run. I suspect the issue is similar to if you were to run the for() loop in my attached code with i set as uint8_t instead of uint16_t. In some cases uint8_t works but in others it can just run on forever because the size of incoming is greater than the max uint8_t value of 255. I really need the ATTINY3226 as the slave due to it's very impressive UART capabilities, but I also really need that final byte. Is there a simple fix to get all 256 bytes?

`#include "Wire.h"

uint8_t incoming[256];

void receiveEvent(int len) {
  for(uint16_t i = 0; i < sizeof(incoming); i++){
    incoming[i] = Wire.read();
    Serial1.print(incoming[i], HEX); Serial1.print(" ");
  }
  Serial1.println();
}

void setup() {
  Wire.begin(0x69);
  Wire.onReceive(receiveEvent); /* register receive event */
  Serial1.begin(921600);
}

void loop() {
  // null
}`

SpenceKonde:
I don't think this is inconsistent with our design intent. I'll make sure the to correct the comment, and ideally #error if someone tries to set it to larger than 255. The issue is the same as the one that caused problems back when the new Wire library went in: The first position in the buffer is occupied by the slave address. We initially failed to account for that, and so the 32-byte buffer was only 31. Since almost every piece of arduino code in the wild takes as a given that the wire buffer is 32 bytes, things that used it at the limit broke; that was quickly corrected once the problem was recognized - but trying to go beyond 255 doesn't seem like an appropriate use of development time. 256 byte I2C payloads seems to be more than is warranted by the 

[...]

PR #840, opened 2022-10-02
Add support for >256 buffers and bugfix

MX682X:
fix and improvement as mentioned in #839

### DIFF
--- a/megaavr/libraries/Wire/src/Wire.cpp
+++ b/megaavr/libraries/Wire/src/Wire.cpp
@@ -346,7 +346,7 @@ uint8_t TwoWire::endTransmission(bool sendStop) {
  */
 size_t TwoWire::write(uint8_t data) {
   uint8_t* txBuffer;
-  twi_buffer_index_t *txHead; 
+  twi_buffer_index_t *txHead;
   #if defined(TWI_MANDS)                   // Add following if host and client are split
     if (vars._bools._toggleStreamFn == 0x01) {
       txHead   = &(vars._bytesToReadWriteS);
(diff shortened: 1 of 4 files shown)


SpenceKonde:
I'm continuing to be confused with the changes here and in DxCore relating to Wire library, because I thought there were additional changes from DxCore waiting to be merged - the development effort is outrunning my ability to keep up with it and know what needs to be merged into where.  Can you let me know what else, if anything, needs to go into mTC and what needs to be ported to DxC for the next release wrt Wire?

MX682X:
Alright, let's see. My PR in DxCore had the main target of adding an example for Dual Mode. But since the github version was not compiling, I decided to change it enough to make it compile. Also, I did some housekeeping in the Wire library itself. Later I realized, I should keep PRs as simple as possible...
Anyway, as far as I know, I also implemented the changes from DxCore to mTC aswell, except the Dual Mode example itself, since the tinies do not support it. Basically, you just have to move the files in the src folder to DxC.

Maybe also useful, a summary of the changes, in chronolgical order:
DxC: added readBytes() function (as known from Serial), removed BusErr enabling and checks. (Due to transmission failure after sleep), made the TWI_checkPins check for scl - 1 == sda (like in Serial)
[PR closed, but those are in mTC]

mTC: add popSleep and pushSleep functions to fix sleep handling, also optimzations to keep the size low. Like: convincing Compiler to use Y-Register in Slave ISR without writing everything in assembly. But this required merging the subfunctions.

mTC: while optimizing, I made a mistake, and this was fixed in this PR.

PR #843, opened 2022-10-07
make sure rxhead never exceeds BUFFER_LENGTH in Wire slave ISR

MX682X:
Shower thought: What happens with the slave if the master is evil or misbehaves and continues to write to the rxBuffer?
Answer: The wire library starts to write outside of itself, aka buffer overflow....

Which is bad. Obviously. So, here is a fix.

If you haven't read my message in the last PR, this is the most up-to-date version, that can be simply copied to DxC.


SpenceKonde:
Yum yum

SpenceKonde:
Manually merged as I'd modified some of wire myself and not committed it. 

I had to move the compile time error checks into Wire.cpp where preictal else they don't every get triggered, (they can still be confused, but it's much better than it was). 

Added method to configure 5 odd options: SDA setup and Hold extension, hold extension on dual mode, and the big one,. SMBus voltage levels, which is a boon those talking with those proliferating 1.8v sensors x_X

PR #869, opened 2022-12-08
Fix ADC1 initialization

mairas:
I am trying to use ADC1 on an ATtiny1616. While debugging my program (in)operation, I realized that ADC1 isn't enabled. I noticed that `init_ADC1()` is defined, but it was referring to ADC0 registers. That's now fixed. Additionally, `init_ADC1()` is now called if the peripheral is present.

This of course isn't a complete implementation of ADC1 support; no `analogReference()` or `analogRead()` implementations are defined. I created my custom implementations for them.

### DIFF
--- a/megaavr/cores/megatinycore/wiring.c
+++ b/megaavr/cores/megatinycore/wiring.c
@@ -1713,25 +1713,25 @@ void __attribute__((weak)) init_ADC0() {
   __attribute__((weak)) void init_ADC1() {
     //                              30 MHz / 32 = 937 kHz,  32 MHz / 32 =  1 MHz.
     #if   F_CPU   > 24000000     // 24 MHz / 16 = 1.5 MHz,  25 MHz / 32 =  780 kHz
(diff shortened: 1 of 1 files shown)


SpenceKonde:
analogRead1() and the like are supplied with the latest version ,or they should be. 

It is intended to not intitialize ADC1 uinless the user code requests it based on my assumption that it is implemented by the compiler as a block of STS instructions, with essentially zero code reuse. Since the percentage of users who will run out of flash is exceeds the percentage who will use the ADC1, calling that function during init() would increase the binary size for far more people who would not want that tradeoff than who will. This is documented in the Analog Reference (or it should be). 

The general use case I had in mind here was that ADC1 would be what people would use for non-standard stuff, like if someone was running one of the ADCs in freerun mode but didn't want to lose the ability to do normal analog read of other pins. But then people started asking about this. 

I noticed the bug with initADC1 just a day or two ago (you don't wanna know how long it's been there for, but that's not really surprising since I didn't add analogRead1() et al until 2.6.x

mairas:
OK, removed the init call.

SpenceKonde:
Actually, maybe I didn't actually fix init_ADC1

PR #889, opened 2023-01-04
add macros to force load pointers into address registers

MX682X:
When analyzing the assembly generated by different gcc versions, I've noticed that there was some potential for optimization by forcing a pointer in the displacement registers.
So, I've created some defines that use the hack I discovered while improving TWI (+ documentation) I'm hoping that this might make things a bit easier in the future. I was able to reduce the core footprint by about 12 (or 14)  bytes on the 1614.


SpenceKonde:
I hope this is a good idea.

PR #895, opened 2023-01-13
Configure "Compile Examples" workflow for authenticated API requests

per1234:
[The `arduino/compile-sketches` GitHub Actions action](https://github.com/arduino/compile-sketches) used in the "Compile Examples" [workflow](https://docs.github.com/actions/using-workflows/about-workflows) queries [the GitHub API](https://docs.github.com/en/rest) for the base ref of the pull request, which is used for the [memory deltas determination](https://github.com/arduino/compile-sketches#enable-deltas-report).

GitHub does [rate limiting of API requests](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting) and if the limit is exceeded, a spurious failure of the workflow run will occur, as was happening previously ([example](https://github.com/SpenceKonde/megaTinyCore/actions/runs/3858167935)).

Authenticated API requests are given a more generous API request allowance, so providing the action with [the automatically generated GitHub access token stored in `secrets.GITHUB_TOKEN`](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) to use for the API requests should prevent any further workflow run failures caused by rate limiting.

### DIFF
--- a/.github/workflows/compile-examples.yml
+++ b/.github/workflows/compile-examples.yml
@@ -229,6 +229,7 @@ jobs:
             -
           verbose: false
           enable-deltas-report: true
+          github-token: ${{ secrets.GITHUB_TOKEN }}
           enable-warnings-report: true
           sketches-report-path: sketches-reports
 
@@ -261,6 +262,7 @@ jobs:
             -
           verbose: false
           enable-deltas-report: true
+          github-token: ${{ secrets.GITHUB_TOKEN }}
           enable-warnings-report: true
           sketches-report-path: sketches-reports
       - name: Compile examples (TCB0 millis)
@@ -291,6 +293,7 @@ jobs:
             -
           verbose: false
           enable-deltas-report: true
+          github-token: ${{ secrets.GITHUB_TOKEN }}
           enable-warnings-report: true
           sketches-report-path: sketches-reports
       - name: Compile examples (RTC millis)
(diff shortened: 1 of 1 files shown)


per1234:
Hmmm. Still failed due to rate limiting. But I see it is authenticated because the error message is different now:

https://github.com/SpenceKonde/megaTinyCore/actions/runs/3910193172/jobs/6682103319#step:5:426

```text
github.GithubException.RateLimitExceededException: 403 {"message": "API rate limit exceeded for installation ID 12198669.", "documentation_url": "https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

vs the error from before the change proposed here:

https://github.com/SpenceKonde/megaTinyCore/actions/runs/3858167935/jobs/6576388511#step:8:423

```text
github.GithubException.RateLimitExceededException: 403 {"message": "API rate limit exceeded for 52.179.102.209. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)", "documentation_url": "https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

The documentation indicates the use of the token provides [1000 requests/hr](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limits-for-requests-from-github-actions), while unauthenticated provides [60/hr](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limits-for-requests-from-github-actions:~:text=60%20requests%20per%20hour). So I'm not sure what is going on.

SpenceKonde:
Hmmm.... Okay, mysterious, but merging

PR #900, opened 2023-01-21
disable gcse

MX682X:
gcse is the thing that "optimizes" Displacements to memory addresses.

First few experiments showed an improvement in code size. I'd like to use the atomated scripts to see how the code size changes throughout all examples. I hope you don't mind, Spence

### DIFF
--- a/megaavr/platform.txt
+++ b/megaavr/platform.txt
@@ -35,12 +35,12 @@ compiler.warning_flags.all=-Wall -Wextra
 # Default "compiler.path" is correct, change only if you want to override the initial value
 compiler.path={runtime.tools.avr-gcc.path}/bin/
 compiler.c.cmd=avr-gcc
-compiler.c.flags=-c -g -Os {compiler.warning_flags} -std=gnu11 -ffunction-sections -fdata-sections -MMD -flto -fno-fat-lto-objects {build.mrelax} -Werror=implicit-function-declaration -Wundef
-compiler.c.elf.flags={compiler.warning_flags} -Os -g -flto -fuse-linker-plugin -Wl,--gc-sections -Wl,--section-start={build.text_section_start} {build.mrelax}
+compiler.c.flags=-c -g -Os {compiler.warning_flags} -std=gnu11 -ffunction-sections -fdata-sections -MMD -flto -fno-fat-lto-objects {build.mrelax} -Werror=implicit-function-declaration -Wundef -fno-gcse
+compiler.c.elf.flags={compiler.warning_flags} -Os -g -flto -fuse-linker-plugin -Wl,--gc-sections -Wl,--section-start={build.text_section_start} {build.mrelax} -fno-gcse
 compiler.c.elf.cmd=avr-gcc
 compiler.S.flags=-c -g -x assembler-with-cpp -flto -MMD
 compiler.cpp.cmd=avr-g++
-compiler.cpp.flags=-c -g -Os {compiler.warning_flags} -std=gnu++17 -fpermissive -Wno-sized-deallocation -fno-exceptions -ffunction-sections -fdata-sections -fno-threadsafe-statics -Wno-error=narrowing -MMD -flto {build.mrelax}
+compiler.cpp.flags=-c -g -Os {compiler.warning_flags} -std=gnu++17 -fpermissive -Wno-sized-deallocation -fno-exceptions -ffunction-sections -fdata-sections -fno-threadsafe-statics -Wno-error=narrowing -MMD -flto {build.mrelax} -fno-gcse
 compiler.ar.cmd=avr-gcc-ar
 compiler.ar.flags=rcs
 compiler.objcopy.cmd=avr-objcopy


SpenceKonde:
Holyfuckballs man... 

The changes are not always negative, but they usually are.... andsomtimes they are huge. Got a link describing what exactlt the thing you disabled was doing such a bad job of?

MX682X:
I've got to admit that changes of up to +/-350 bytes are certainly unexpected. I was afraid that disabling one of the optimizations would increase code-size, but I certainly did not expect this span. (just saw your response...)

It's hard to explain. In order to find the culprit I've asked gcc to dump everything ( add -fdump-tree-all -fopt-info -fdump-rtl-all  if you want to take a look, this generates a lot of files) and I noticed that the "cprop" pass was the pass that created the problems. Also called Copy Propagation. It was squashing the REG+offset insn. It was also doing more stuff, but before looking through the other code, I thought it would be worth trying to disable it first. To see what happens

SpenceKonde:
```
Compiling sketch: megaavr/libraries/tinyNeoPixel/examples/RGBWstrandtest
Compilation time elapsed: 1s
Change in flash: -2556 (-15.6%)
Change in RAM for global variables: -14 (-1.37%)
```

*blink blink* dafuk....

MX682X:
It's extremely weird. I know at least that I'm looking at right spot, I just gotta figure out how the thing works....

SpenceKonde:
Okay, what is this sorcery?! On my system I don't see the savings....

MX682X:
I've noticed it once before. Someone had opened an issue and they had code (lst file) that was different from my system (Windows). They were using a Mac. If I recall correctly, you were using a Mac too, right? I guess the GCC binaries are different. For whatever reason....

SpenceKonde:
A Mac? What kind of bourgoise do you think am? Have you seen me wearing
outfits made from hjundred dollar bills sewn together with platinum group
metals?! Of course I don't have a Mac, I can't afford that shit!!

But, from the ci we know x64 linux seems seems to share this issue based pm
the CC. I run windows because I'm too cheap for Mac and want a real
computer, rulling out anything with an apple on it

___________
Spence Konde
Azzy’S Electronics

Contact: ***@***.***

MX682X:
oops, sorry,. I'm not touching Apple stuff voluntarily either, so I feel that.

But back to the topic, I don't think adding this is a good idea - especially since I finally started to understand how the middle and back ends of gcc work. I'm actually trying to get my own optimization pass running. My plan is now to see if I can achieve *something* with a winows build. If it works, I'll try to get the other platforms to compile.
(I wish I had better things to keep my brain busy...)

SpenceKonde:
Note that the failing checks were spurious and due to rate limiting not actual failures

SpenceKonde:
Okay, the behavior here is just whack, like, it's ALL OVER THE PLACE. Changing the clock source sometimes changes size by 60 bytes or more. Othertimes not at all. And so on. And it is intensely diffciult to analyze because there doesn't seem to be any way to get the size results out of the CI runs automatically, and manually selecting the text, because it's loaded dynamically, takes several minutes per test case. 

While the net impact is positive, nearly as many test cases are worse than those that are better in the one test case I ran the numbers for. 

What a load of crap, avr-gcc's optimization is terribad

MX682X:
Due to the chaotic results, I had the feeling that it would be non reliable. That's why I didn't want to pursue disabling this optimization pass. 

And the only reason why avr-gcc's optimizations are whack, is because the support is lacking. And it is lacking, because microchip is trying to sell their own compiler (-03 and -Os are behind a paywall). 
But I was able to achieve some success in my odysee: I got the gas-interrupts pseudo instructions backported (Assembler analyzes ISR code and reduces the push/pop overhead to whatever is actually needed).

While I'm at it, I was thinking about a speed optimization for 32 bit + const:
```
(load low word in r24, 25)
adiw r24 0x01
(store)
brcc +6 (Branch if carry clear, skip the high byte)
(load high word in r24, r25)
adiw r26 0x01
(store)
```
in the worst case, this would add 1 clock (+6%)(once every 256 times), in the best reduce it by 6 clocks (-37,5%) 
Of course this only works with non volatile variables and in ISRs (like millis)

SpenceKonde:
I think there's a typo in there, I don't see how that achieves what you
want.

Oh, I hope you don't mind that I'm renaming all the "dirty trick" type
macros. They getting put In to camelcase like functions, with no
underscores except one at the start, and in come cases the words are
getting changed too. They are macros I will end up using (I've considered
doing this anyway).

I'm moving all of them into a new file in the core called dirty_tricks.h.
having an underscore keeps it clear thst they are not supposed to be for
general use. but they were physically painful to type in all caps, and they
made the code look godawful. They care intensely about the types of the
variables thst are passed (or rather what the types need to be for them to
not be a disaster - they won't error if they get bunk types, they'll do
something more evil) so they're more like a function than a typical macro,
in terms of how they have to be used. Bur the real reason is just I plan to
use them, and I can't stand all these long allcaps names, they hurt to
type, and microchip already imposes way too many way too long constants in
allcaps.

I probably wouldn't do that for an explicitly public api implemented with
the expectation that they are generally useful, but this is the trunk where
I'm stuffing all the ugly duct tape we use for performance.

I'm also documenting it.....

____________
Spence Konde
Azzy’S Electronics

Contact: ***@***.***

MX682X:
> Oh, I hope you don't mind that I'm renaming all the "dirty trick" type
> macros. They getting put In to camelcase like functions, with no
> underscores except one at the start, and in come cases the words are
> getting changed too. They are macros I will end up using (I've considered
> doing this anyway).
> 

not at all. Documentation was one of the reasons I wanted to put the macros in one spot in the first place....

MX682X:
@SpenceKonde 
https://www.dropbox.com/s/2c0j2wy9kgchrjv/avr-gcc-7.3.0-atmel3.6.1-mxduino14-x86_64-w64-mingw32.zip?dl=0
https://www.dropbox.com/s/4toewv0u2561p8z/avr-gcc-7.3.0-atmel3.6.1-mxduino14-x86_64-pc-linux-gnu.tar.bz2?dl=0
Well... finally finished the alpha version of the memory analyzer and displacement generator. Also, the win version is 32-bit.
Haven't tested it well yet, in fact only some results of lst files. Of course, the core reductions aren't big - as most improvements were already implemented in assembly.
Also had to learn that gcc does not assign addresses to variables, the assembler does that, so it's impossible to analyze multiple variables, only structs.
In order to use it, you need to add '-mfavor-displacement' to the flags

PR #937, opened 2023-04-19
ATtiny814 - Fix PWM output for WO0 & WO2

brunob45:
### Context
I wanted to build the [Adafruit Seesaw Peripheral](https://github.com/adafruit/Adafruit_seesawPeripheral) project on an ATtiny814, but the PWM function of PB3 was not working; it's the alternate PIN for WO0.

I forked the repo and saw major changes since the official release of 1.6.5 in this area. This looked promising.

### Changes
I fixed some copy-paste errors and added a check in `analogWrite` and `digitalWrite` to handle this special case.

### Side note
I'm also thinking of adding another lookup table describing the relationof `pin number` to `TCA output`. As of right now, it is computed, but adding alternate pin functionality may complicate this code too much:
```c
uint8_t offset = 0;
if (bit_mask > 0x04) { // HCMP
  bit_mask <<= 1;      // mind the gap
  offset = 1;          // if it's an hcmp, the offset of the compare register is 1 higher.
}
if      (bit_mask & 0x44) offset += 4;
else if (bit_mask & 0x22) offset += 2;
timer_cmp_out = ((volatile uint8_t *)(&TCA0.SPLIT.LCMP0)) + (offset); //finally at the very end we get the actual pointer (since volatile variables should be treated like nuclear waste due to performance impact)
(*timer_cmp_out) = (val); // write to it - and we're done with it.
TCA0.SPLIT.CTRLB |= bit_mask;
```

I could make another PR with this change if the proposed solution is acceptable.

### DIFF
--- a/megaavr/cores/megatinycore/wiring_analog.c
+++ b/megaavr/cores/megatinycore/wiring_analog.c
@@ -1133,9 +1133,13 @@ void analogWrite(uint8_t pin, int val) {
         #else
           //Otherwise, we're in split mode and we use the classical method.
           volatile uint8_t *timer_cmp_out; // must be volatile for this to be safe.
-          #ifdef __AVR_ATtinyxy2__
-            if (bit_mask == 0x80) {
-              bit_mask = 1;  // on the xy2, WO0 is on PA7
+          #if defined(_TCA_ALT_WO0)
+            if (bit_mask == PIN3_bm) {
+              bit_mask = PIN0_bm;
+            }
+          #elif defined(__AVR_ATtinyxy2__)
+            if (bit_mask == PIN7_bm) {
+              bit_mask = PIN0_bm;  // on the xy2, WO0 is on PA7
             }
           #endif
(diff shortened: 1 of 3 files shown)


SpenceKonde:
This is literally what I am working on now. I've got 3 of the 4 half-banks of TCA pins working, buffered mode appears to work, only thing I need to get sorted is why certain combinations are not being used, but I think this may a boards.txt issue or something (there are an absurd number of options for the 24-pin parts), and check the much smaller numbers of options on smaller parts.

This is being treated as very serious because it turns out that I never released 2.6.6, and 2.6.5 couldn't burn bootloader, and both had critical bugfixes in them. By the time I realized it, I had disassembled the PWM machinery

brunob45:
Thanks! Should I close this PR then?

I'm available to test PWM functions on the tiny814 or 1607 if you want.

SpenceKonde:
github version should now work

PR #946, opened 2023-04-24
Fix PWM output alternate pin for xy4 variant

brunob45:
Fix `copy paste` error in `digital_pin_to_timer` array


SpenceKonde:
Thanks! Right you are!

PR #954, opened 2023-05-11
Update dirty_tricks.h

MX682X:
Ok so, there may or may not have been a case where I accidentely used a wrong pointer for this macro.... so I thought, it would be a good idea to make it less dirty... badArg wasn't shown on my end if I supplied a wrong type, but it still didn't let me compile and I think it's always better to have some save guard. Better save then sorry, y'know?

### DIFF
--- a/megaavr/cores/megatinycore/dirty_tricks.h
+++ b/megaavr/cores/megatinycore/dirty_tricks.h
@@ -258,11 +258,12 @@ Not enabled. Ugly ways to get delays at very small flash cost.
 /************/
 /* _fastPtr */
 /************/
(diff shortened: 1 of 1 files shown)


SpenceKonde:
Cool - that checking is all compiletime right? Merging. We're getting mad failures right now but that's because I don't know github yaml well enough to fix it. I may just rename the internal name of some options on the 2-series have to reference the mux of a non-existant TCD (in the name - not any of the compile flags that come with it or anything, but otherwise I need to figure out how to give different names to the 0, 1 and 2 series PWMmap options when running the test

PR #975, opened 2023-06-19
Add option to enable XDIR pin for RS485 mode

xieliwei:
I've added an option to enable RS485 mode 1 on the UART so that the XDIR pin can be used to control an external RS485 transceiver. Unlikely to be useful if you have more than one device on the network, but great for remotely flashing hard-to-access devices that you would otherwise need to physically get to.

With a suitable USB RS485 cable that automatically drives the bus without fancy RTS signaling, everything works (avrdude/Arduino upload) without any additional modification. You will of course need to figure out how to enter the bootloader remotely.

This change increases the bootloader size by an additional instruction to set the XDIR pin to output. For BIGBOOT, it also embeds the RS485 parameter.


SpenceKonde:
confirmed generated binaries under normal circumstances are identical

Issue #452, opened 2021-06-14
Mystery: EEPROM issues when accessing EEPROM immediately after startup. 


yuranikspb:
```
void setup() {
// EEPROM.write(55, 55) 
//delay(2);
flag_first = EEPROM.read(40);

  if (flag_first != 17)
  {
    EEPROM.write(40, 17); //not work if no delay  2 ms or  EEPROM.write(random adress, random byte)
  }  
```

delay(1) not help
delay(2) work good
if  on start do EEPROM.write(random adress, random byte) work good

bod level ON
bootloader No

yuranikspb:
i m test with attiny 412 this sketch  if not use delay, EEPROM.write(40, 17); will never work
```
#include <EEPROM.h>
byte flag_first=0;

void setup() {
//delay(2)
flag_first = EEPROM.read(40);

pinMode(4,OUTPUT);

if (flag_first != 17)
  {
    EEPROM.write(40, 17);
    digitalWrite(4,1);
   
  }
else{

digitalWrite(4,0);
}

}
```

SpenceKonde:
I uploaded this to latest released of megaTinyCore, to a tiny412, The first time I uploaded, the light went on. As I would expect, Power cycled, led did not go on. Uploaded again, light did not go on. Changed number  (was 17 like you had, changed to 33). Ligt went on first time as it wrote the new value, and not after that. 

Note that, obviously when I set the fuses on the board I set them to retain EEPROM (you must do burn bootloader to apply that setting as is documented in readme - by default , a chip erase (which is performed every time code is uploaded through UPDI) will erase the contents of the EEPROM - unless EESAVE is is set to retain (not the default setting on a new part). 

```
#include <EEPROM.h>
byte flag_first=0;
void setup() {
//delay(2)
flag_first = EEPROM.read(40);
pinMode(1,OUTPUT);
if (flag_first != 33)
  {
    EEPROM.write(40, 33);
    digitalWrite(1,1);  
  }
else{
digitalWrite(1,0);
}
}
void loop() {
}
```
When I do this,

SpenceKonde:
Can you provide a method to reproduce this consistently, and describe how you prove that the EEPROM is not being written? Also, ensure that you've set the EESAVE fuse correctly

SpenceKonde:
Also what BOD level, and what kind of sort of powert supply at what voltage?

SpenceKonde:
Also when it fails, - what are the symptoms?

[...]

PR #986, opened 2023-07-15
Update parse_options.mk

rneurink:
Pass the entry conditions to the GCC compiler. These are currently not passed to the GCC compiler. See Issue on the dxCore as well [Issue #452](https://github.com/SpenceKonde/DxCore/pull/452)


SpenceKonde:
thaaank you.

PR #1055, opened 2024-02-08
Update README.md

avandalen:
This text was confusing and pointed to itself, completely unnecessary


HeadBoffin:
Sort of - if you open it up from your hard disk it could be out of date - perhaps change it to say "See the latest here ..."

SpenceKonde:
Thanks, I have clarified that sentence. The reason for having that there at all is that they could just open the .md file in notepad, and that is not a wonderful user experience.

PR #1071, opened 2024-03-05
Create package.json

mechatroniks-git:
Allows for easy platform.io platform_packages to be used. 

See https://community.platformio.org/t/upgrading-individual-platform-packages/31731/4

### DIFF
--- a/megaavr/package.json
+++ b/megaavr/package.json
@@ -1,0 +1,15 @@
+{
+  "name": "framework-arduino-megaavr-megatinycore",
+  "version": "2.6.10",
+  "description": "Arduino Wiring-based Framework for Microchip tinyAVR 0-series and 1-series chips (megaTinyCore)",
+  "keywords": [
+    "framework",
+    "arduino",
+    "microchip",
+    "tinyAVR"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/SpenceKonde/megaTinyCore"
+  }
+}


SpenceKonde:
Why is this pointing towards your git? 

Please link to where the platform.json is documented so that I can construct a correct one that does not rely on external resources.

I can't move ahead with anything that relies on an personal account that is not my own, as in the future you could be compromised and the content replaced with malicious content, which I would then be tarred and feathered for. Or you could vanish and I could be flooded with support inquiries asking when it was fixed and could only say "Don't know, I don't know what he did, and he vanished", since I don't know shit about pio (which I gather this is for - I was looking into how to add this file myself a few days ago.

SpenceKonde:
Does this file get deleted from the board manager install packages or not? 


(when I do a board manager release, I use a script that (as of more recently than the last release was actually done) pulls the repo, uncompresses it, deletes the gargantuan uncompressable image files from megaavr/extras (which is also where the documentation is, while leaving in place compact image files (just based on file type, as it happens that the gigantic files I don't want in the board manager downloads all have the same file extension and similar names), moves everything up a folder level because Arduino needs that for board manager, then performs some crude edits to the platform.txt (it changes the path of the SerialUPDI upload to where board manager installs it, rather than where a sane person would manually place it. Repo points towards the sane location for development with manual installation, whereas the releases have that corrected.)


Is there a way to point them to the github zip file of the most recent release? It is not unusual for the repo to spend a long time between releases like, totally busted as I tackle a major change, (which was started intentionally only after doing a release, since I knew it was going to potentially hose things for a while). Automatic installs are meant to just work, so I only want to use actual releases for that (hell, my track record with those is nothing to write home about, except maybe to complain, but it's miles better than the repo itself.),

SpenceKonde:
Can you please clarify these points? I know there are a lot of PIO people who would like a feature like this, I hate getting questions from them that I can't answer so anything that will make fewer people complain about installation problems, and more people use a version from the modern era, instead of versions of the core that are so old that I was writing the documentation in hieroglyphics, that's a good thing. The PIO users are either extremely numerous, or the world's squeakiest wheels.

mechatroniks-git:
Apologies for switching to my git, was trying to pull it from there. It's been reverted.

To the best of my knowledge, all we need to do is include the (corrected) package.json so that platform.io can pull it from github, as opposed to 2.6.7 which they have in their package. Then the platformio user brings in the github version using [platform packages](https://docs.platformio.org/en/latest/projectconf/sections/env/options/platform/platform_packages.html).

I don't think it needs to be in the boards manager install package, as platform.io doesn't really use anything from Arduino. 

platform_packages looks like it can pull from a commit or from a zip, an example I found [here](https://community.platformio.org/t/platform-packages-not-installing-when-using-git-reference/27048):
`platform = https://github.com/tasmota/platform-espressif32/releases/download/v2.0.3rc1/platform-espressif32-2.0.3.zip`

or a [release](https://community.platformio.org/t/using-specific-version-of-esp-idf-and-arduino-esp32-causes-in-a-include-error/21155/10)
`platform_packages = framework-arduinoespressif32@https://github.com/espressif/arduino-esp32.git#idf-release/v3.3`

SpenceKonde:
Great, thanks. It will be done on all cores from the next release.

coupla questions

So, SerialUPDI as you may know has a dependance on python (specifically python3); it's one of the "tools" that we install when installed via arduino ide board manager). How should I deal with that dependance for SerialUPDI? Just document that python3 must be in the path if they want to use SerialUPDI? (that means separate release file for PIO, of course, as the other two cases will have it in a specific place given as the full path after Arduino does it's substitution on platform.txt) Rip out SerialUPDI entirely? 

Also, what is the directory structure supposed to be? ("the same as Arduino" does not answer this question, because it does not uniquely identify a directory structure, because it's different when installed via board manager and manually.... 

Do you happen to have a paragraph I could insert into the installation documentation to explain to to the platformio'ers how that json file gets used and what they need to do?

mechatroniks-git:
Yeah, I can create an example structure and docs for this. But essentially, it is a straightforward platformio project, add libraries and modify the `platformio.ini` file something like this below - which references the same python that Arduino calls. 

```
; PlatformIO Project Configuration File
;
;   Build options: build flags, source filter
;   Upload options: custom upload port, speed and extra flags
;   Library options: dependencies, extra library storages
;   Advanced options: extra scripting
;
; Please visit documentation for the other options and examples
; https://docs.platformio.org/page/projectconf.html

[env:ATtiny1624]
platform = atmelmegaavr
framework = arduino
board = ATtiny3227
board_hardware.oscillator = internal
board_build.f_cpu = 10000000L
board_hardware.bod = disabled
board_hardware.eesave = yes
board_hardware.rstpin = updi
upload_protocol = custom
upload_speed = 230400
upload_port = /dev/cu.SLAB_USBtoUART
build_flags = -DMILLIS_USE_TIMERB0=1
upload_flags =
	--tool
	uart
	--uart
	$UPLOAD_PORT
	--clk
	$UPLOAD_SPEED
	--device
	$BOARD
upload_command = 
C:/Users/winpc/AppData/Local/Arduino15/packages/megaTinyCore/hardware/megaavr/2.5.11/python -u C:/Users/winpc/AppData/Local/Arduino15/packages/megaTinyCore/hardware/megaavr/2.5.11/tools/libs/pymcuprog/prog.py $UPLOAD_FLAGS -wd 1 --fuses 0:0b00000000 2:0x02 6:0x04 7:0x00 8:0x00 -f $SOURCE -a write -v
```

In this file, I'll also add in a platform_packages call to override the 2.6.7 megatinycore that platformio still uses.

SpenceKonde:
Well how will we know if python is present?l amd where? We only know on
windows systems because we tell board manager to install this standalone
package Co training pyhton..... What if there is no python installed on the
system? What if it is the wrong version?

Sorry for all the questions s and confusion. I don't k ow jack shit about
platform io other Tham thst they users are loud and complainy
____________
Spence Konde
Azzy’S Electronics

Contact: ***@***.***

Issue #1103, opened 2024-05-23
Change some variables in tinyNeoPixel from private to protected


moose4lord:
Hi, I'm using a NeoPixel library that inherits from the Adafruit_NeoPixel class and would like to modify the library to instead inherit from your tinyNeoPixel class. Adafruit_NeoPixel exposes some of the nitty gritty details of its inner workings by declaring some critical variables as "protected", making them available to classes that inherit from Adafruit_Neopixel.
```c++
protected:
#ifdef NEO_KHZ400 // If 400 KHz NeoPixel support enabled...
  bool is800KHz; ///< true if 800 KHz pixels
#endif
  bool begun;         ///< true if begin() previously called
  uint16_t numLEDs;   ///< Number of RGB LEDs in strip
  uint16_t numBytes;  ///< Size of 'pixels' buffer below
  int16_t pin;        ///< Output pin number (-1 if not yet set)
  uint8_t brightness; ///< Strip brightness 0-255 (stored as +1)
  uint8_t *pixels;    ///< Holds LED color values (3 or 4 bytes each)
  uint8_t rOffset;    ///< Red index within each 3- or 4-byte pixel
  uint8_t gOffset;    ///< Index of green byte
  uint8_t bOffset;    ///< Index of blue byte
  uint8_t wOffset;    ///< Index of white (==rOffset if no white)
  uint32_t endTime;   ///< Latch timing reference
```

Unfortunately, those variables are declared private in tinyNeoPixel, so are not reachable. Would you be willing to change the declaration to protected?

Thanks for sharing all your ATtiny hard work.

SpenceKonde:
Can you explain what the reasoning is behind:
1. Why they might ever have been made private? I am a C person, this C++ object oriented stuff and classes - I don't have the instinct there.

moose4lord:
I have to go back to 2017 to find a version of the Adafruit_Neopixel lib that has those variables declared as private. When Adafruit released v1.1.2 of that lib on Jun 28, 2017, they changed the designation from private to protected. No real mention of it in the release, they just kind of snuck it in there.

The three access specifiers in C++ are:
public - members are accessible from outside the class.
private - members cannot be accessed from outside the class. 
protected - members cannot be accessed from outside the class, however, they can be accessed in inherited classes.

[...]

PR #1114, opened 2024-06-16
changed tinyNeoPixel lib variables from private to protected

moose4lord:
Oops, deleted my original pull request by accident, so resubmitting. See issue #1103.


moose4lord:
Thanks Spence

PR #1121, opened 2024-07-06
Update PlatformIO.md serialUPDI instructions

Elon61:
Added a bit more information in regards to using SerialUPDI in PlatformIO since i struggled a little getting it to work. 

The custom method was inspired by [Henry Cheung](https://www.e-tinkers.com/2023/03/make-megatinycore-work-for-platformio/)'s blog post, and allows you to use this core's uploader with very little hassle, but i would suggest simply using the `serialupdi` upload protocol which worked well enough for my purposes.


SpenceKonde:
Thanks

PR #1134, opened 2024-08-07
Fix pin assignment to use TUNING_PIN define

mgrosvenor:
Currently the `TUNING_PIN` pin define is unused, and the `pulseIn()` function uses a hard coded value,


SpenceKonde:
oops

SpenceKonde:
oops again on the errant commit referencing this. Was supposed to be #1143.

PR #1164, opened 2024-11-05
Fix hang in Serial.flush()

jvasileff:
_state will be 0x2 for half-duplex mode, so we need to be careful with operator precedence to test only the low bit. Without this change, Serial.flush() hangs when half-duplex is enabled and no data has yet been sent.


SpenceKonde:
That's not the only time It would break. Jeez. I know better than that! I'm disappointed with me.

PR #1167, opened 2024-11-16
Fix time gap with RTC millis() overflow

jvasileff:
The millis state was internally represented with a 32 bit number, using timer_overflow_count as the high 16 bits, and RTC.CNT as the low 16 bits. The max value before overflow was 0xFFFFFFFF (or 4294967295) which corresponded to about 4194303999ms given the timer's frequency of 1024Hz. But then, one millisecond later, after overflow, the internal value 0x0 would correspond to 0ms, skipping all millisecond values between (inclusive) 4194304000 and 2^32-1, causing an apparent time warp into the future by 100663296ms or 28 hours.

After this commit, the internal representation is in milliseconds, which results in no time warp.

----

To test, the following code can be used, which both checks for gaps in the returned value of `millis()` and the accuracy of `millis()` over repeating intervals.

```c
#include <Arduino.h>

const int INTERVAL_MS = 2000; // must be < 4195 for 16MHz clock
const int TOLLERANCE_MS = INTERVAL_MS / 20 + 1;

uint32_t lastIntervalMillis = 0;

void startTimerA0() {
    // raise TCA_SINGLE_OVF_bm every INTERVAL_MS

    uint16_t top = (F_CPU / 1000) * INTERVAL_MS / 1024; // Assuming a prescaler of 1024
    TCA0.SINGLE.CTRLA = 0; // Stop the timer
    TCA0.SINGLE.CTRLD = 0; // Ensure the timer is not in split mode
    TCA0.SINGLE.PER = top; // Set the period value
    TCA0.SINGLE.CNT = 0; // Initialize the count to 0
    TCA0.SINGLE.INTFLAGS = TCA_SINGLE_OVF_bm; // Clear the overflow flag
    TCA0.SINGLE.CTRLA = TCA_SINGLE_CLKSEL_DIV1024_gc | TCA_SINGLE_ENABLE_bm; // Set the prescaler and enable the timer
}

void setup() {
    Serial.begin(115200);
    Serial.println("Starting up");
    Serial.flush();

    startTimerA0();
    lastIntervalMillis = millis();
}

boolean timerTriggered() {
    if (TCA0.SINGLE.INTFLAGS & TCA_SINGLE_OVF_bm) {
        TCA0.SINGLE.INTFLAGS = TCA_SINGLE_OVF_bm;
        return true;
    }
    return false;
}

void loop() {
    static uint32_t lastMessageMillis = millis() / 1000UL * 1000UL;
    static uint32_t lastLoopMillis = millis();
    uint32_t currentLoopMillis = millis();

    if (currentLoopMillis - lastMessageMillis >= 1000) {
        Serial.printf("%lums: Running, all ok\n", currentLoopMillis);
        lastMessageMillis += 1000;
    }

    boolean error = false;

    if (timerTriggered()) {
        uint32_t elapsed = currentLoopMillis - lastIntervalMillis;
        lastIntervalMillis = currentLoopMillis;
        if ((elapsed > INTERVAL_MS + TOLLERANCE_MS)
                || (elapsed < INTERVAL_MS - TOLLERANCE_MS)) {
            Serial.println("Error: elapsed per millis() does not match interval");
            Serial.printf("    elapsed: %lu, interval: %i\n", elapsed, INTERVAL_MS);
            error = true;
        }
    }

    if (currentLoopMillis - lastLoopMillis > 1) {
        Serial.println("Error: millis() jumped");
        error = true;
    }

    if (error) {
        Serial.printf("    lastLoopMillis:    %lu\n", lastLoopMillis);
        Serial.printf("    currentLoopMillis: %lu\n", currentLoopMillis);
        Serial.println("halting.");
        while (true);
    }

    lastLoopMillis = currentLoopMillis;
}
```

Since the rollover problem only occurs after about 48 days, the following patch can be applied to `wiring.c` to queue rollover in 64 seconds:

```diff
diff --git i/megaavr/cores/megatinycore/wiring.c w/megaavr/cores/megatinycore/wiring.c
index 4fa43ae4..55ef9ecf 100644
--- i/megaavr/cores/megatinycore/wiring.c
+++ w/megaavr/cores/megatinycore/wiring.c
@@ -1431,6 +1431,7 @@ void __attribute__((weak)) init_millis()
       pTCD->INTCTRL        = 0x01; // enable interrupt
       pTCD->CTRLA          = TIMERD0_PRESCALER | 0x01; // set clock source and enable!
     #elif defined(MILLIS_USE_TIMERRTC)
+      timingStruct.timer_overflow_count = 0xffff;
       while(RTC.STATUS); // if RTC is currently busy, spin until it's not.
       // to do: add support for RTC timer initialization
       RTC.PER             = 0xFFFF;
```

With this, running the test program produces the following output:

```
...
4194300000ms: Running, all ok
4194301000ms: Running, all ok
4194302000ms: Running, all ok
4194303000ms: Running, all ok
0ms: Running, all ok
Error: millis() jumped
    lastLoopMillis:    4194303999
    currentLoopMillis: 0
halting.
```

After this PR is applied, the following can be used to queue rollover to verify the fix:

```diff
diff --git i/megaavr/cores/megatinycore/wiring.c w/megaavr/cores/megatinycore/wiring.c
index 3cd8ae59..7d09aaab 100644
--- i/megaavr/cores/megatinycore/wiring.c
+++ w/megaavr/cores/megatinycore/wiring.c
@@ -1391,6 +1391,7 @@ void __attribute__((weak)) init_millis()
       pTCD->INTCTRL        = 0x01; // enable interrupt
       pTCD->CTRLA          = TIMERD0_PRESCALER | 0x01; // set clock source and enable!
     #elif defined(MILLIS_USE_TIMERRTC)
+      timingStruct.timer_millis = 0xffffffffUL - 10000UL; // start 10 seconds before overflow
       while(RTC.STATUS); // if RTC is currently busy, spin until it's not.
       // to do: add support for RTC timer initialization
       RTC.PER             = 0xFFFF;
```

and the test program produces the following output:

```
...
4294964000ms: Running, all ok
4294965000ms: Running, all ok
4294966000ms: Running, all ok
4294967000ms: Running, all ok
704ms: Running, all ok
1704ms: Running, all ok
2704ms: Running, all ok
3704ms: Running, all ok
...
```

### DIFF
--- a/megaavr/cores/megatinycore/wiring.c
+++ b/megaavr/cores/megatinycore/wiring.c
@@ -171,12 +171,8 @@ uint8_t __PeripheralControl = 0xFF;
   #define MILLIS_INC (millisClockCyclesToMicroseconds(TIME_TRACKING_CYCLES_PER_OVF)/1000)
 
   struct sTimeMillis {
-    #if (defined(MILLIS_USE_TIMERB0) || defined(MILLIS_USE_TIMERB1) || defined(MILLIS_USE_TIMERB2) || defined(MILLIS_USE_TIMERB3) || defined(MILLIS_USE_TIMERB4))     // Now TCB as millis source does not need fraction
+    #if (defined(MILLIS_USE_TIMERRTC) || defined(MILLIS_USE_TIMERB0) || defined(MILLIS_USE_TIMERB1) || defined(MILLIS_USE_TIMERB2) || defined(MILLIS_USE_TIMERB3) || defined(MILLIS_USE_TIMERB4))     // Now TCB as millis source does not need fraction
       volatile uint32_t timer_millis;   // That's all we need to track here
-
-    #elif defined(MILLIS_USE_TIMERRTC)  // RTC
-      volatile uint16_t timer_overflow_count;
-
     #else                               // TCAx or TCD0
       //volatile uint16_t timer_fract;
       //volatile uint32_t timer_millis;
@@ -191,10 +187,8 @@ uint8_t __PeripheralControl = 0xFF;
   // Now for the ISRs. This gets a little bit more interesting now...
   #if defined (MILLIS_USE_TIMERRTC)
     ISR(MILLIS_TIMER_VECTOR) {
-      // if RTC is used as timer, we only increment the overflow count
-      // Overflow count isn't used for TCB's
       if (RTC.INTFLAGS & RTC_OVF_bm) {
-        timingStruct.timer_overflow_count++;
+        timingStruct.timer_millis += 64000;
       }
       RTC.INTFLAGS = RTC_OVF_bm | RTC_CMP_bm; // clear flag
     }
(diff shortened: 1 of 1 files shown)


SpenceKonde:
That is a nasty bug you found. Time travel is not permitted! Especially not by 28 hours, that's long enough to buy a winning lotto ticket.

PR #1215, opened 2025-03-13
Fix typos relating to TCD usage, document WOC and WOD

pdw-mb:
Fixes a couple of (what I think are) typos in comments relating to WOC and WOD, and also attempts to document these outputs on the TCD0 page.

### DIFF
--- a/megaavr/extras/Ref_TCD.md
+++ b/megaavr/extras/Ref_TCD.md
@@ -11,7 +11,7 @@ In addition to the synchronization, some of the intentional features make it cha
 ```c++
 analogWrite(PIN_PA4,128); // 50% PA4. - like usual
 analogWrite(PIN_PA7,192); // 50% PA4, 75% PA7 - like usual
-analogWrite(PIN_PA6, 64); // 25% PA5, 25% PA6, 75% PA7 - PA4 and PA6 are both channel A
+analogWrite(PIN_PA6, 64); // 25% PA4, 25% PA6, 75% PA7 - PA4 and PA6 are both channel A
 //use digitalWrite to turn off the PWM on a pin
 // or call turnOffPwm(pin);
 ```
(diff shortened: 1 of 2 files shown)


megaavr/extras/Ref_TCD.md
```diff
@@ -16,6 +16,10 @@ analogWrite(PIN_PA6, 64); // 25% PA4, 25% PA6, 75% PA7 - PA4 and PA6 are both ch
 // or call turnOffPwm(pin);
 ```
 
+## WOC and WOD (PC0 and PC1)
+
+On 20/24 pin 1-Series parts, `TCD0` can also drive the `WOC` and `WOD` pins (PC0 and PC1 respectively). `WOC` and `WOD` can each be connected to either `WOA` or `WOB`, but the core only supports connecting `WOC` to `WOA` and `WOD` to `WOB`.
```
SpenceKonde, line 21: The changes in this document do not represent the state of the functionality presently available in megaTinyCore. 

We actually use WOC and WOD by default, because we run TCA0 in splitmode (analogWrite is 8-bit) so the WOA and WOB don't become intereresting until pins PC4 and PC5 become available (24 pinonly), at which point you can switch WO5 and WO4 to those pins, and use PA4/PA5 for TCD. In github version the tools PWM pins menu should work for setting the pins to get PWM out of. (there were some bugs in current release)

SpenceKonde:
First two are accurate. 

Second is accurate, but not reflective of the API presented by the core or how the core configures those timers by default.

SpenceKonde:
There are considerable factual inaccuracies that need to be fixed in those documents - A whole huge section of the TCD reference is a copy-paste error and describes a feature that would take too much space on most tinies, and is only available on the Dx-seriesvia dxcore

PR #1252, opened 2025-10-28
Fix logical condition for UART buffer sizes in UART0.cpp

smilexs4:


smilexs4:
No problem. I guess it's fairly uncommon to need a UART buffer that large; hence, no one reported the compilation error. My application reads a lot of Modbus registers, which needs the large buffer.

PR #1270, opened 2026-05-10
Extension to include debugging support 

felias-fogg:
This PR adds support for debugging provided by [PyAvrOCD](https://pyavrocd.io) to megaTinyCore. In addition to this PR, you need to include the tool package pyavrocd in the boards manager index file. See the end of https://felias-fogg.github.io/megaTinyCore/package_SpenceKonde_megaTinyCore_index.json for an example.

### DIFF
--- a/megaavr/platform.txt
+++ b/megaavr/platform.txt
@@ -16,7 +16,7 @@ version={versionnum.major}.{versionnum.minor}.{versionnum.patch}{versionnum.post
 # Stupid workaround #
 # for IDE bug       #
 #####################
-version=2.6.11
+version=2.7.0-pre1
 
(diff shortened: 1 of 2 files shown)


felias-fogg:
BTW: If you would like to try it out, you can add https://felias-fogg.github.io/megaTinyCore/package_SpenceKonde_megaTinyCore_index.json to the Additional boards manager URLs and then install "megaTinyCore (Debug enabled)." Plugging in a Curiosity Nano board and firing up Arduino IDE 2.3.8 should give you a first impression.

felias-fogg:
Oops. I noted that a number of the fuses should also be set when uploading the executable. Since there is no provision yet in pyavrocd for doing so, we should wait with including the extension until this is implemented.

